PR #11, opened 2019-02-16
Fix integer overflow when determine system memory size

alexey-lysiuk:
On macOS sysconf(_SC_PAGE_SIZE) returns 4096 and sysconf(_SC_PHYS_PAGES) returns 4'194'304
scpagesiz * scphyspages is equal to 0x4'0000'0000 which becomes zero after casting to 32-bit unsigned integer
This leads to 'BUFFER TOO BIG TO FIT IN CACHE!' fatal error on startup


nukeykt:
Thanks

PR #21, opened 2019-02-18
Update .gitignore with entries for MSVC and Xcode

alexey-lysiuk:
Built Debug/Release configurations and excluded all untracked files
Fixed mix of Windows/UNIX line endings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,10 @@ apps/
 *.cfg
 textures
 
+/platform/Windows/Win32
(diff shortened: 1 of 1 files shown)


alexey-lysiuk:
BTW I didn't add game data files to `.gitignore` like `platform/Windows/*.art`, etc. Probably, it's worth adding them as well.

nukeykt:
Thanks, merged

PR #23, opened 2019-02-19
Remove luajit from dependencies in Travis configuration

alexey-lysiuk:
Apparently, `luajit` isn't used at the moment, it was added by mistake.


nukeykt:
Merged. Thanks 😃

Issue #8, opened 2019-02-15
Status of macOS target


alexey-lysiuk:
I'm curious about status of macOS target, and of Xcode project in particular. Will you plan to update it?
Probably you don't own a Mac because it's a bit outdated and tightly coupled with Homebrew for some reason.
I would like to fix this. However, I'm in doubt about dependencies. Is it OK to have them in the repo?

nukeykt:
I tried to build EDuke32 last week on macos, it compiled fine using makefile, though it crashed due to overflow in Bgetsysmemsize function. After fixing overflow it run fine. Xcode project is definitely outdated, yeah. As for NBlood i never tried to build it.
Yes. Feel free to add deps in repo.

alexey-lysiuk:
OK, I already configured EDuke and Mapster. 

Regarding `Bgetsysmemsize()` please take a look at [this fragment](https://github.com/nukeykt/NBlood/blob/8307879aaa7216d16c556081e0862fac83f39512/source/build/src/compat.cpp#L608-L616).

`sysconf(_SC_PAGE_SIZE)` returns 4096 which is expected.
`sysconf(_SC_PHYS_PAGES)` returns 4'194'304 which is OK as well.

Although, `scpagesiz * scphyspages` is 17'179'869'184.
It overflows `uint32_t` and becomes zero because `0x4'0000'0000 & 0xFFFF'FFFF == 0`.

Do not really how this should work.

nukeykt:
IIRC i changed it to something like `siz = (uint32_t)min<uint64_t>(UINT32_MAX, scpagesiz * scphyspages);` in order to make it working.

alexey-lysiuk:
I have a question about iOS in Xcode project. I removed related targets in process of project cleanup.
The thing is I have zero interest in it. Probably, I should ask you first about your plans regarding iOS.

nukeykt:
I have no plans to port it to iOS, so it's fine

alexey-lysiuk:
Merged in 407b568df1817780ab74245b47202adb88137bfe.

PR #29, opened 2019-02-24
Xcode macOS build environment

alexey-lysiuk:
Like we discussed in #8 I finished update of Xcode project to make macOS version much easier to develop.
As a side note, we can avoid addition of headers and libraries to repository. However, this will require running a script to download and build them.


nukeykt:
Thanks again 😃 Hope it works fine. I haven't tested it on macOS by myself, but i hope to try it soon as possible.

alexey-lysiuk:
I tested on a fresh clone using Xcode 10.1 and with the exception of a few compilation warnings I'm not aware of any issues.

PR #41, opened 2019-03-22
fix compiling without opengl

Jan200101:
when compiling without OpenGL tileProcessGLVoxels is called yet never defined nor declared


nukeykt:
Thanks

PR #45, opened 2019-04-03
Big endian fixes

BSzili:
Tested on big endian Linux. Currently the Post Mortem demo desyncs, but I plan to investigate it later.


nukeykt:
Nice, thanks. Post Mortem demo desync is normal behavior, it desyncs in original game as well

PR #64, opened 2019-05-11
Fix building via Xcode project

alexey-lysiuk:
`aiunicult.cpp` and `aiunicult.h` were missing from Xcode project


nukeykt:
Great, thanks!

PR #65, opened 2019-05-11
Fix AppVeyor builds

alexey-lysiuk:
Switched to VS2019 Preview image as there is no release version provided yet
Still much better than builds erroneously reported as broken


nukeykt:
Nice, didn't know that it had VS2019 preview image. Thanks again 😄

PR #74, opened 2019-06-02
GetOptions fixes

Phlamethrower:
It was impossible to use the -game_dir option, or use absolute paths on Unix due to confusion with Windows "/" switch character.

I think there are still a few quirks in the options handling to resolve, but this should get rid of a few of the bigger ones.


nukeykt:
Looks good, merging. Thanks!

PR #75, opened 2019-06-02
Fix tileLoadTile to check if tile already loaded

Phlamethrower:
While working on my RISC OS port I spotted that there was a lot of stuttering, especially when firing weapons. Some debugging revealed that it was because tileLoadTile wasn't checking to see if the tile was already loaded before calling tileLoad.

### DIFF
--- a/source/blood/src/tile.cpp
+++ b/source/blood/src/tile.cpp
@@ -147,7 +147,7 @@ void tileProcessGLVoxels(void)
 
 char * tileLoadTile(int nTile)
 {
-    tileLoad(nTile);
+    if (!waloff[nTile]) tileLoad(nTile);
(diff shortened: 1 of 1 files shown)


nukeykt:
Merging, thanks

Rinnegatamante:
@Phlamethrower related to this issue. After your PR, do you have huge stuttering issues when the second demo pops in after the first one finished in main menu? Your PR actually solved stuttering issues in my PSVITA port too in game however that second demo completely kills framerate on Vita (getting like 1 new frame every 3-4 seconds).

Phlamethrower:
Yeah, that level (e3m2) does seem pretty laggy. I'll see if I can take a closer look next time I do some work on the project - hopefully sometime in the next week or two.

PR #83, opened 2019-07-16
Fix linux compilation

Jan200101:
Various things in the code did not allow for compilation under Linux
I think this PR resolves that

Changes:
- _MAX_PATH defines as PATH_MAX
- Window renamed to Win, the X header already typedefs Window as an XID
- replace itoa with snprintf, should give the same results

This PR needs to be tested with MSVC to ensure nothing broke on the Windows side

### DIFF
--- a/source/blood/src/db.cpp
+++ b/source/blood/src/db.cpp
@@ -34,6 +34,11 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #include "db.h"
 #include "iob.h"
 
+#ifdef __linux__
+#include <linux/limits.h>
+#define _MAX_PATH PATH_MAX
+#endif
+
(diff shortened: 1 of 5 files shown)


Jan200101:
Hmmm… odd
The Linux CI builds fail because of issues I do not encounter on my personal machine

I am going to investigate this further and fix that too

Jan200101:
Linux CI works now
the OSX CI is still broken because of a missing _MAX_PATH declaration
I am thinking about defining it to PATH_MAX too

nukeykt:
Thanks.
I think we should replace _MAX_PATH with BMAX_PATH here.

Jan200101:
Everything except the XCODE CI works now
sadly fixing that part is out of my reach, see #84

nukeykt:
Ok, thanks. Merging.

PR #88, opened 2019-08-07
Fix the crosshair

CommonLoon102:
Currently the crosshair looks different in NBlood than in the original game. In the original it consists of multiple colors like the corresponding tile but in NBlood it is just one color. It is because NBlood is recoloring the crosshair by the "crosshaircolor" console feature, although nothing is asking for it.
I kept the possibility to change the color of the crosshair from the console or config file. I also added the "crosshairreset" console command if somebody wants to restore the original crosshair, so it is not needed to quit the game, edit the config file (by deleting the crosshaircolor setting) and restart the game.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,5 @@ DerivedData/
 *.dSYM/
 
 .DS_Store
+/texturecache
+/texturecache.index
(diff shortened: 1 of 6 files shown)


nukeykt:
Thanks, merged

Issue #26, opened 2019-02-22
Do not restart music after loading a level with the same track


alexey-lysiuk:
Currently, when loading a saved game music playback starts from the beginning.
It would be great if music will continue to play if a loaded game uses the same track.

CommonLoon102:
I'm on this issue, I have a fix already, I'll introduce a setting for it because the music restarts in the original game too.

CommonLoon102:
Fixed in master.

PR #89, opened 2019-08-08
Add an option to let the music play continuously without restart if…

CommonLoon102:
…the player loads a game with the same map the player is already playing on.

On higher difficulties it is often very annoying that the music restarts from the beginning every time the player loads a saved game for the same map. It causes that only the first few seconds are played over and over again. In case of CD audio it is even worse because some of the tracks are starting with silence or low volume.

The original game had this behavior too, so I've kept it as the default, and I've added a new audio setting for this feature. BloodGDX has this feature too but it is enabled by default and cannot be turned off. My philosophy is that every new feature or modification should be opt-in to keep the game as close to the original as  possible.

Related issue: https://github.com/nukeykt/NBlood/issues/26

### DIFF
--- a/source/blood/src/blood.cpp
+++ b/source/blood/src/blood.cpp
@@ -99,6 +99,9 @@ bool bNoAutoLoad = false;
 
 bool bVanilla = false;
 
+int gMusicPrevLoadedEpisode = -1;
+int gMusicPrevLoadedLevel = -1;
+
 char gUserMapFilename[BMAX_PATH];
 char gPName[MAXPLAYERNAME];
 
@@ -533,6 +536,8 @@ void StartLevel(GAMEOPTIONS *gameOptions)
     EndLevel();
     gStartNewGame = 0;
     ready2send = 0;
+    gMusicPrevLoadedEpisode = gGameOptions.nEpisode;
+    gMusicPrevLoadedLevel = gGameOptions.nLevel;
     if (gDemo.at0 && gGameStarted)
         gDemo.Close();
     netWaitForEveryone(0);
--- a/source/blood/src/blood.cpp
+++ b/source/blood/src/blood.cpp
@@ -467,7 +467,7 @@ void PreloadCache(void)
     if (gDemo.at1)
         return;
     if (MusicRestartsOnLoadToggle)
-        sndPlaySpecialMusicOrNothing(MUS_LOADING);
+        sndTryPlaySpecialMusic(MUS_LOADING);
     gSoundRes.PrecacheSounds();
     PreloadTiles();
     int clock = totalclock;
--- a/source/blood/src/blood.cpp
+++ b/source/blood/src/blood.cpp
@@ -463,7 +463,8 @@ void PreloadCache(void)
     char tempbuf[128];
     if (gDemo.at1)
         return;
-    sndPlaySpecialMusicOrNothing(MUS_LOADING);
+    if (MusicRestartsOnLoadToggle)
+        sndPlaySpecialMusicOrNothing(MUS_LOADING);
     gSoundRes.PrecacheSounds();
     PreloadTiles();
     int clock = totalclock;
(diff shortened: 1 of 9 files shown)


CommonLoon102:
Do not merge this yet, I found some bugs related to changing maps, in that case no music is played, I'll have to work on this a little more.

CommonLoon102:
E1M5 tricked me as it doesn't have any music defined in the blood.ini, so it was OK. But I found another bug: save the game before ending a level, end the level and go to the next, load the save. In this case the second level's music was played instead of the first one. This is now fixed. Also, now if a demo was played the game will restart the music anyways, even if the loaded map is the same what was played in the demo.

source/blood/src/blood.cpp
```diff
@@ -463,7 +466,8 @@ void PreloadCache(void)
     char tempbuf[128];
     if (gDemo.at1)
         return;
-    sndPlaySpecialMusicOrNothing(MUS_LOADING);
+    if (MusicRestartsOnLoadToggle)
+        sndPlaySpecialMusicOrNothing(MUS_LOADING);
```
nukeykt: I think you can use sndTryPlaySpecialMusic here instead

source/blood/src/config.cpp
```diff
@@ -322,6 +323,7 @@ void CONFIG_SetDefaults(void)
     MouseBias       = 0;
     MouseDeadZone   = 0;
     MusicToggle     = 1;
+    MusicRestartsOnLoadToggle = 1;
```
nukeykt: Default it to 0

source/blood/src/loadsave.cpp
```diff
@@ -91,7 +91,7 @@ void LoadSave::Write(void *pData, int nSize)
         ThrowError("File error #%d writing save file.", errno);
 }
 
-void LoadSave::LoadGame(char *pzFile)
+void LoadSave::LoadGame(char *pzFile, bool demoWasPlayed)
```
nukeykt: You could move `gDemo.at1` and `gDemo.Close()` here from `void LoadGame(CGameMenuItemZEditBitmap *pItem, CGameMenuEvent *event)` instead

source/blood/src/menu.cpp
```diff
@@ -569,8 +570,9 @@ const char *pzMusicDeviceStrings[] = {
 };
 
 CGameMenuItemTitle itemOptionsSoundTitle("SOUND SETUP", 1, 160, 20, 2038);
-CGameMenuItemZBool itemOptionsSoundSoundToggle("SOUND:", 3, 66, 60, 180, false, UpdateSoundToggle, NULL, NULL);
-CGameMenuItemZBool itemOptionsSoundMusicToggle("MUSIC:", 3, 66, 70, 180, false, UpdateMusicToggle, NULL, NULL);
+CGameMenuItemZBool itemOptionsSoundSoundToggle("SOUND:", 3, 66, 50, 180, false, UpdateSoundToggle, NULL, NULL);
+CGameMenuItemZBool itemOptionsSoundMusicToggle("MUSIC:", 3, 66, 60, 180, false, UpdateMusicToggle, NULL, NULL);
+CGameMenuItemZBool itemOptionsSoundMusicRestartsOnLoadToggle("MUSIC RESTARTS:", 3, 66, 70, 180, false, UpdateMusicRestartsOnLoadToggle, NULL, NULL);
```
nukeykt: I don't think this option is worth exposing to menu

source/blood/src/config.cpp
CommonLoon102: That would be different from the original game, are you sure you want to make behavioral changes by default compared the the original? Are there already things what differs from the original on purpose already?

source/blood/src/menu.cpp
CommonLoon102: If you are sure that you want to change by default the original behavior then I agree, the console command would be enough for 99.99% of the players.

source/blood/src/config.cpp
nukeykt: i'm fine with it, this is not chocolate port anyway. In long run it'll be EDuke32 equivalent for blood

nukeykt:
Thanks

Issue #67, opened 2019-05-14
GUI cursor is difficult to see


nashmuhandes:
I use the mouse to navigate the menus and the cursor is hard to see, initially. I have some suggestions on how the experience might be improved:

A) Dim the 3D viewport while the menu is up (like GZDoom). Would default to off to replicate vanilla visuals.

B) Ability to customize the mouse cursor graphic. I feel like it hides into the surroundings too much because it re-uses the in-game crosshair as the cursor.

C) Always place the mouse cursor at the center of the screen when the UI is first opened. This is my biggest issue... when you are playing the game using the mouse, the cursor moves with your mouselook actions in the game, even if the UI isn't opened. So every time I re-open the menu, the cursor would be at some random position and I'll have to move the mouse around first to find the cursor before I can even begin to operate the menu.

CommonLoon102:
I've placed the mouse cursor to the middle when the menu is first opened, I've also replaced the cursor to a rotated bloody arm gib (tile 3521), I might decorate it with dripping blood or something and then let's see what @nukeykt thinks about it.

CommonLoon102:
Fixed in master.

PopeRigby:
@CommonLoon102 What happened to the bloody arm? It's an arrow in master.

CommonLoon102:
Bloody arm idea has been dropped, didn't look that good.

Issue #92, opened 2019-08-09
Power-up countdown


CommonLoon102:
There could be a power-up countdown similar to Fresh Supply.
The question is what about the multiplayer because if somebody doesn't turn this on that will be a disadvantage for them. Maybe it should be turned on by default for everybody and if somebody doesn't like it, it could be disabled from the console (like the no-restart music option).

PR #93, opened 2019-08-11
 Mouse cursor improvements in menu and powerup display on the HUD

CommonLoon102:
- Move the mouse cursor to the center of the screen when opening up the menu
- Change the mouse cursor to something else for better visibility (also it was sometimes confusing why there are two aim reticles)
- Remove focus from the item what was under the mouse cursor if up or down keys were pressed

- Show the active power-ups and their remaining time on the HUD

Related issues:
https://github.com/nukeykt/NBlood/issues/67
https://github.com/nukeykt/NBlood/issues/92

### DIFF
--- a/source/blood/src/gamemenu.cpp
+++ b/source/blood/src/gamemenu.cpp
@@ -107,7 +107,10 @@ bool CGameMenuMgr::Push(CGameMenu *pMenu, int nItem)
         m_mouselastactivity = -M_MOUSETIMEOUT;
         m_mousewake_watchpoint = 0;
         mouseLockToWindow(0);
-        MoveMouseToCenter();
+        if (m_bFirstPush)
+            m_bFirstPush = false;
+        else
+            mouseMoveToCenter();
     }
     dassert(pMenu != NULL);
     if (nMenuPointer == 8)
@@ -122,16 +125,6 @@ bool CGameMenuMgr::Push(CGameMenu *pMenu, int nItem)
     m_menuchange_watchpoint = 1;
     m_mousecaught = 1;
     return true;
-}
-
-void CGameMenuMgr::MoveMouseToCenter()
-{
-    SDL_Window* window = SDL_GL_GetCurrentWindow();
-    SDL_GetWindowSize(window, &m_windowSize.x, &m_windowSize.y);
-    if (m_bFirstPush)
-        m_bFirstPush = false;
-    else
-        SDL_WarpMouseInWindow(window, m_windowSize.x / 2, m_windowSize.y / 2);
 }
 
 void CGameMenuMgr::Pop(void)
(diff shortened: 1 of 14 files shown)


source/blood/src/gamemenu.cpp
```diff
@@ -122,6 +124,16 @@ bool CGameMenuMgr::Push(CGameMenu *pMenu, int nItem)
     return true;
 }
 
+void CGameMenuMgr::MoveMouseToCenter()
+{
```
nukeykt: I don't think using SDL directly here is a good idea, either move this to the engine or make this code active only for SDL builds

source/blood/src/view.cpp
```diff
@@ -1158,6 +1158,66 @@ void viewDrawStats(PLAYER *pPlayer, int x, int y)
...
+                POWERUPDISPLAY temp = powerups[j];
+                powerups[j] = powerups[j+1];
+                powerups[j+1] = temp;
+            }
+        }
+    }
+}
+
+void viewDrawPowerUps(PLAYER* pPlayer)
+{
```
nukeykt: Add a toggle for this in menu

source/blood/src/gamemenu.h
```diff
@@ -471,14 +472,17 @@ class CGameMenuMgr
     int32_t m_mousecaught;
     vec2_t m_prevmousepos, m_mousepos, m_mousedownpos;
     bool m_postPop;
+    vec2_t m_windowSize;
```
nukeykt: Use engine's xdim and ydim instead

source/blood/src/gamemenu.cpp
```diff
@@ -182,25 +194,64 @@ void CGameMenuMgr::Draw(void)
 
         m_mousewake_watchpoint = 0;
     }
+
+    vec2_t cursorpos = { (m_mousepos.x >> 16) + 9, (m_mousepos.y >> 16) + 18};
```
nukeykt: Shifting coordinate by 16 makes cursor position locked at 320x200 resolution coordinates

source/blood/src/gamemenu.cpp
```diff
@@ -182,25 +194,64 @@ void CGameMenuMgr::Draw(void)
...
+
     // Display the mouse cursor, except on touch devices.
-    if (MOUSEACTIVECONDITION)
+    if (MOUSEACTIVECONDITION && !m_bFirstPush)
     {
-        int32_t a = kCrosshairTile;
+        int16_t mousetile = 3522; // disembodied arm
```
nukeykt: Leave crosshair variant for now, i'm going to add more cursor variants later

source/blood/src/gamemenu.cpp
```diff
@@ -182,25 +194,64 @@ void CGameMenuMgr::Draw(void)
...
+            rotatesprite_fs_alpha(cursorpos.x << 16, cursorpos.y << 16, z, 0, mousetile, 0, NULL, stat, alpha);
         }
     }
     else
         g_mouseClickState = MOUSE_IDLE;
+
+    DrawBloodDripFromCursor(cursorpos);
+}
+
+void CGameMenuMgr::DrawBloodDripFromCursor(const vec2_t& cursorpos)
```
nukeykt: I'm not fan of this idea, i'd remove it

source/blood/src/view.cpp
CommonLoon102: Sure, it takes no time, I just wonder if it is needed at all? Why would anybody turn this off? It is like the audio restarting feature which is off now by default and that doesn't have a toggle either.

CommonLoon102: Another thing what I was thinking about is the default value of the toggle. Should this be on or off be default? Because if it is off by default and somebody doesn't know about this setting then it would be a handicap for somebody in multiplayer.
The remaining time and the sprite for a power up is only visible when the player has picked up such power-up, and it doesn't cover out much of the screen.
If you still want a toggle, sure, what should be the default value of it?

source/blood/src/gamemenu.cpp
CommonLoon102: What about tile 1043 (red arrow pointing up)? And rotate a little so it would look similar as in BloodGDX, just maybe not that big and different color. Just an idea.

source/blood/src/gamemenu.cpp
CommonLoon102: I'm just shifting it back, a few lines above I was shifting it the other direction.

source/blood/src/gamemenu.cpp
nukeykt: I like this variant

source/blood/src/view.cpp
nukeykt: Enable it by default. Music restarting thing is somewhat very subtle change imo, so cvar was enough there

source/blood/src/gamemenu.cpp
CommonLoon102: By "this" you mean the red arrow or the aim reticle?

nukeykt: red arrow

nukeykt:
Thanks!

CommonLoon102:
fixes #67

Issue #91, opened 2019-08-08
Items replenish on map change


CommonLoon102:
All items except the med-kit replenish when you go from a map to the next. In the HUD the correct previous values will be shown, but when you start using them they will count down from 100 (shown from 99).

This is a bug also in the original 1.21, maybe these kind of bugs shouldn't be fixed, as it is not the fault of the port. I think if we would fix this we would have to fix the bug where you can kill large gray gargoyles by using crouch too. Or maybe we could have a new setting group for bug fixes and people could opt-in as they like?

PR #97, opened 2019-08-13
 bugfix: do not replenish items on next map start 

CommonLoon102:
Related issue: https://github.com/nukeykt/NBlood/issues/91

This is a trivial bug in the original DOS version (v1.21) too. When you win a map and the game loads the next map then you will have your items replenished (except the med-kit). If you look at the inventory, it will show the correct percentage, but when you use an item the first time on the new map, it will replenish and count down from 100%.

Do you think this should be fixed in NBlood?

### DIFF
--- a/source/blood/src/player.cpp
+++ b/source/blood/src/player.cpp
@@ -821,6 +821,26 @@ void playerResetInertia(PLAYER *pPlayer)
     viewBackupView(pPlayer->at57);
 }
 
+void playerResetPowerups(PLAYER* pPlayer)
+{
+    const int jumpBoots = 15;
+    const int divingSuit = 18;
+    const int crystalBall = 21;
+    const int beastVision = 25;
+
+    for (int i = 0; i < kMaxPowerUps; i++)
+    {
+        if (!VanillaMode()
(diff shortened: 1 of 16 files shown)


CommonLoon102:
Maybe use `VanillaMode()` here too?

source/blood/src/player.cpp
```diff
@@ -942,6 +947,12 @@ void playerStart(int nPlayer)
     pPlayer->nWaterPal = 0;
     for (int i = 0; i < kMaxPowerUps; i++)
         pPlayer->at202[i] = 0;
```
nukeykt: I guess this would be better:
```
    for (int i = 0; i < kMaxPowerUps; i++)
    {
        if (i == 15 || i == 18 || i == 21 || i == 25)
            continue;
        pPlayer->at202[i] = 0;
    }
```

nukeykt:
> Maybe use `VanillaMode()` here too?

Yeah

source/blood/src/player.cpp
CommonLoon102: I agree.

Issue #101, opened 2019-08-14
funky shoes cheat code


CommonLoon102:
The current behavior is the same as in DOS 1.21 but it is strange.
1. If you first type it it gives you the jump boots.
2. The second time it just deactivates them. Although, it says "You have no Jumping Boots."

The problem with case 1 is that if you already have jump boots it just activates them but won't max it out to 100% before activating it.
And with the 2nd case is that you still have the boots despite it says you don't.

Issue #100, opened 2019-08-13
Clarice cheat code does nothing


CommonLoon102:
It should give you 100% armor for all 3 kind of armors. This is a bug in the original v1.21 too. I guess it worked for older versions.

Issue #98, opened 2019-08-13
Cheesehead and cousteau cheats


CommonLoon102:
They give you a diving suit, but if your diving suit wasn't completely zeroed before, then they (the cheat codes) don't really give you a diving suit, because as soon as you start using it, it will go back to the previous level. If you had 40% before entering the cheat then it will fall from 100% to 40% instantly when you start using it. 

This bug is also in the original DOS version (v1.21).

Issue #90, opened 2019-08-08
Calgon cheat does not work


CommonLoon102:
It should jump to the next level, instead it does nothing.
If you use it like the mario cheat then it will work like mario, e.g.: calgon 2 3

This bug is also in the original 1.21 game.
I think what happened is that they added the mario cheat and created a bug because they didn't handle calgon anymore correctly.

Since it is a bug in the original too, I tend to think that we should leave it as it is.

PR #102, opened 2019-08-14
Fixing some cheat code related issues

CommonLoon102:
All of these bugs are in the original v1.21 too, but I think they should be fixed.

### LARA CROFT:
Related issue: https://github.com/nukeykt/NBlood/issues/14
What I did is that I kept the pitchfork after taking all the weapons. I've done like this because the FORK YOU cheat code does the same when it takes away all the weapons to not freeze the game on weapon switch. This not just solves the issue for LARA CROFT but it fixes also the `weapons` console command too as they do the same eventually.

### CALGON:
Related issue: https://github.com/nukeykt/NBlood/issues/90
In older versions of Blood it jumped to the next level, although in version 1.21 it is just an alias for MARIO. So, I think it should work without parameters, in this case it should load the next map. With parameters, it should work like MARIO. 

This is how it works now with this patch.

### CHEESEHEAD and COUSTEAU
Related issue: https://github.com/nukeykt/NBlood/issues/98
Now they work as they intended to work. Only the percent indicator is updated to 100% but the available amount for the item was not updated by the cheat codes. It is a similar bug like this: https://github.com/nukeykt/NBlood/issues/91

### CLARICE
Related issue: https://github.com/nukeykt/NBlood/issues/100
It should give you 100% armor for all 3 kind, instead it did nothing, I've added the missing code.

### FUNKY SHOES
Related issue: https://github.com/nukeykt/NBlood/issues/101
Now it will really take away your shoes when you type it the second time.
It will also replenish the shoes to 100% if you already have them.

### VOORHEES, ONERING, JOJO, TEQUILA
Related issue: https://github.com/nukeykt/NBlood/issues/104
When it says your are not invulnerable/invisible/delirious anymore, now you really won't be.


nukeykt:
Can you bound these changes to non vanilla mode using `VanillaMode()` check? (currently only used only for demos, but i'll eventually add toggle to menu)

CommonLoon102:
> 
> 
> Can you limit these changes to non vanilla mode using `VanillaMode()` check? (currently only used only for demos, but i'll eventually add toggle to menu)

Sure, I will do!

nukeykt:
Thanks

PR #105, opened 2019-08-14
Krueger armor fix

CommonLoon102:
Typing the KRUEGER cheat sets you on fire and prints out FLAME RETARDANT, although you will still die. If you are flame retardant, you shouldn't die, right? It seems they have assigned the wrong value to the variable, they wanted to give you 200 fire armor so they assigned 200, but that means about 12 fire armor, and it is not enough and you will die. Changing 200 to 3200 will result in 200 fire armor and you will survive the fire. I've added the VanillaMode() guard.


source/blood/src/messages.h
```diff
@@ -82,43 +82,43 @@ class CCheatMgr
...
-        kCheat30,
-        kCheat31,
-        kCheat32,
-        kCheat33,
-        kCheat34,
-        kCheat35,
-        kCheat36,
-        kCheat37,
+        kCheat1, // refills ammo, no cheat code for it
+        GRISWOLD,
```
nukeykt: Please use kCheat prefix and use uppercase only for first char

Issue #107, opened 2019-08-15
Zombies can catch fire underwater


CommonLoon102:
This is how it is in DOS vanilla v1.21. It just strange that you can ignite the zombies underwater, in Fresh Supply you can't.

PR #110, opened 2019-08-15
No burning under water

CommonLoon102:
Related issue: https://github.com/nukeykt/NBlood/issues/107

I think it makes no sense to let zombies became flameguys under water, so this patch won't let them.
Also, if you shot cultists with the flare gun secondary, they became flameguys for a very short time and then they are converting back. The patch fixes this too.

Another fix: if you shot a wall from close with the flare gun secondary you caught fire underwater and burnt for long, now this is not possible.
Also, enemies won't be flaming (not the same as the flameguy) underwater, just for a very short amount of time.


nukeykt:
Sorry, but i'm against of significant gameplay changes at this point

Issue #108, opened 2019-08-15
Tiny Calebs turn into large flameguys


CommonLoon102:
On E6M2 when breaking the mirror the little Calebs will come out. If you set them on fire with the spray can they will turn into large flameguys. This is also vanilla behavior. Shouldn't they stay little when they are flaming?

CommonLoon102:
I think it would need a new SEQ with tiny flameguy, seems like out of scope.

CommonLoon102:
It is possible to scale SEQs.

CommonLoon102:
Scaling changed in mapedit branch, tiny Caleb scale factor needs to be changed when merging with master: https://github.com/nukeykt/NBlood/commit/c5d28a2f09f7603956e2cb77fe17539fb0dbee55

CommonLoon102:
Scale factor changed to 64 in 70c1e0ed

PR #113, opened 2019-08-16
Fix tiny Calebs turning into large flameguys

CommonLoon102:
Related issue: https://github.com/nukeykt/NBlood/issues/108

This is just a cosmetic fix, doesn't affect gameplay.
The other code change where I fixed the AI state pointer also doesn't change anything since the structs have the same value.


CommonLoon102:
When https://github.com/nukeykt/NBlood/commit/c5d28a2f09f7603956e2cb77fe17539fb0dbee55 gets merged into master, -4 needs to be changed to 64.

nukeykt:
Their `BurnSearch` state are not the same though!

CommonLoon102:
> 
> 
> Their `BurnSearch` state are not the same though!

How come? Sorry, I don't what you mean.

nukeykt:
`AISTATE cultistBurnSearch = { kAiStateSearch, 3, -1, 3600, NULL, aiMoveForward, thinkSearch, &cultistBurnSearch };`
`AISTATE innocentBurnSearch = { kAiStateSearch, 3, -1, 3600, NULL, aiMoveForward, thinkSearch, NULL };`
Next state is differs

CommonLoon102:
> 
> 
> `AISTATE cultistBurnSearch = { kAiStateSearch, 3, -1, 3600, NULL, aiMoveForward, thinkSearch, &cultistBurnSearch };`
> `AISTATE innocentBurnSearch = { kAiStateSearch, 3, -1, 3600, NULL, aiMoveForward, thinkSearch, NULL };`
> Next state is differs

OK, I'm blind :)

nukeykt:
Wouldn't setting dude type to 252 be sufficient (which has proper x/yrepeat in SEQ files)?

CommonLoon102:
> 
> 
> Wouldn't setting dude type to 252 be sufficient (which has proper x/yrepeat in SEQ files)?

If I set `pSprite->type = 252;` instead of scaling down the SEQ then they will behave differently, they will run around while flaming. If the type is 239 then they will stand still while flaming.

CommonLoon102:
I have reverted every change which is not related to the size issue. By the way those changes didn't change any behavior, or I didn't notice.
I think you need to cherry pick the SEQ resizing change from mapedit branch before we want to merge this, so we won't forget to change -4 to 64 later.

For the behavior changes and issues, I've opened a separate issue, it is easier if we handle the size problem and the behavior problems separately because they are indeed a separate problem.
Maybe you decide that changing the behavior for flaming little Calebs and innocents are too huge gameplay changes for this mod, then we don't need to even start working on it and then I will close this issue: https://github.com/nukeykt/NBlood/issues/120

PR #115, opened 2019-08-16
Show map title at the beginning of the maps

CommonLoon102:
Nowadays the map loading screen is just a flash and it is not possible to read the name of the next level. Therefore, now it shows the map name for 2 seconds at the beginning of the level, fading out after 1 second. It is using the same position and style as the loading screen. So the text is the same as on the loading screen, it looks like it is from there for 2 more seconds.


source/blood/src/view.cpp
```diff
@@ -1226,6 +1228,28 @@ void viewDrawPowerUps(PLAYER* pPlayer)
...
+        return;
+
+    const int noAlphaForSecs = 1;
+    uint8_t alpha = videoGetRenderMode() != REND_CLASSIC || numalphatabs >= 15 ?
+        seconds < noAlphaForSecs ? 0 : clamp(((seconds-noAlphaForSecs)*1000+millisecs)/4, 0, 255)
+        : 0;
+
+    if (alpha != 255)
+    {
+        viewDrawText(1, pzLoadingScreenText2, 160, 50, -128, 0, 1, 1, 0, alpha);
```
nukeykt: Use levelGetTitle() instead of pzLoadingScreenText2

nukeykt:
Thanks

PR #116, opened 2019-08-17
fix that map title was shown in every minute

CommonLoon102:
I should have tested it a little more.


source/blood/src/view.cpp
```diff
@@ -1234,7 +1234,7 @@ void viewDrawMapTitle(void)
     int minutes = gLevelTime / (kTicsPerSec * 60);
     int seconds = (gLevelTime / kTicsPerSec) % 60;
     int millisecs = (gLevelTime % kTicsPerSec) * 33;
-    if (seconds > 3)
+    if (minutes != 0 || seconds > 3)
```
nukeykt: why not remove minutes completely and not modulo seconds?

PR #118, opened 2019-08-17
 show the author and difficulty at 2D map mode 

CommonLoon102:
After you have started a new game you cannot see anywhere what difficulty you are playing on. Also, I've added to display the author of the map, because why not.


nukeykt:
I don't think this would be necessary after savegame menu updating, which is i'm planning to do down the road

PR #123, opened 2019-08-18
 Add Fresh Supply (Steam) to game data search paths 

CommonLoon102:
I don't have the GOG version of Fresh Supply so I couldn't add it.


nukeykt:
Thanks

Issue #68, opened 2019-05-14
Feature request: Map sky texture to the interior of a cylinder


01y:
Currently, the sky is rendered as a scrolling plane with an orientation that depends on the where the player is looking. This produces an odd effect when looking up, which does not correspond to any physical scenario I can think of. Engines such as GZDoom solve the problem of translating a scrolling sky texture into geometrically correct 3d by mapping the texture to the interior of a cylinder, and filling the remaining space with a solid color, depending on the sky texture, which blends in well with the texture's upper edge, with very good results. Requesting that NBlood/ EDuke32 do the same.

CommonLoon102:
Yes, we need a skybox like in the original DOS version. Currently, it looks different.

CommonLoon102:
Strangely, it does something on the Cryptic Passage maps, because it looks different than the base game maps.
Polymost in E1M1:
[Imgur](https://imgur.com/dTPPdaZ)

Polymost in CP4 (double Moons):
[Imgur](https://imgur.com/qBlWutX)

nukeykt:
Polymost has hack to tile sky texture if it has 256px or less y resolution. It's used for Duke 3d's cloud sky texture

CommonLoon102:
If somebody wants real skyboxes instead of the glitchy one in polymost mode what is now in the game, you can use this nblood.pk3 instead of the original one: https://github.com/CommonLoon102/NBlood/blob/skybox/nblood.pk3

CommonLoon102:
It got fixed in 45d9bb0a and 2298fe16 so this issue could be closed.

PR #124, opened 2019-08-18
Skybox fix

CommonLoon102:
This is fixing the skyboxes in polymost.
Related issue: https://github.com/nukeykt/NBlood/issues/68


nukeykt:
I'm not willing to include skyboxes in distribution. Also stuff like palswaps/shading will not work with true-color images and what about if some mod will replace stock sky tiles?

PR #127, opened 2019-08-19
fix Illegal instruction crash on 32 bit AMD processors and Intel Pentium 3

CommonLoon102:
In the make file a few lines above it sets the architecture to Pentium 3 and then later it adds the SSE2 instruction set optimizations. Pentium 3 doesn't have SSE2, nor any 32 bit AMD CPU. Only Pentium 4 has it. It caused Illegal instruction crash for me on my AMD Athlon XP.
The problematic line was at source/build/src/osd.cpp:90
It wants to assign a double value and the compiler generated SSE2 code, only SSE2 has double float operators, which the Pentium 3 doesn't have.

### DIFF
--- a/Common.mak
+++ b/Common.mak
@@ -597,26 +597,25 @@ ifndef OPTOPT
         ifeq ($(PLATFORM),DARWIN)
             OPTOPT := -march=nocona -mmmx -msse -msse2 -msse3
         else
-            OPTOPT := -march=pentium3
+            OPTOPT := -march=pentium4
             ifneq (0,$(GCC_PREREQ_4))
                 OPTOPT += -mtune=generic
                 # -mstackrealign
             endif
-            OPTOPT += -mmmx -msse -mfpmath=sse
+            OPTOPT += -mmmx -msse -msse2 -mfpmath=sse
+
+            # Fix for older PCs than Pentium 4
+            ifeq ($(HOSTPLATFORM),LINUX)                
+                ifneq ($(shell $(CC) -march=native  -dM -E - < /dev/null | grep -i "__SSE2__" | wc -l),1)
+                    OPTOPT := -march=native
+                endif
+            endif
+
         endif
     endif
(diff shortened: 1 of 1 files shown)


nukeykt:
Enabling SSE/SSE2 was a fix polymost renderer glitches in 32 bit builds, so disabling it is a bad idea

CommonLoon102:
SSE/SSE2? Which one? SSE or SSE2? They are different things. If the polymost renderer requires SSE2 then it means that there will be only one 32 bit CPU which will be able to run the game, which is the Pentium 4. Then why pentium3 is written in the make file?

If polymost requires SSE2 then the minimum requirement for NBlood would be a Pentium 4 and then it couldn't work on any other 32 bit CPU (talking about x86 of course)!

CommonLoon102:
One more thing, what if I don't want to use polymost, but only software renderer? Currently you cannot run the game, it won't even start, not even in software renderer because of SSE2.

nukeykt:
Probably eduke32 devs forgot to change target architecture

CommonLoon102:
What if I detect SSE2 from the makefile and then turn off USE_OPENGL if it is not present? Also not appending -msse2 in that case of course.
That way CPUs without SSE2 will have software rendering only, but at least you can play the game.
What do you think?

CommonLoon102:
I didn't mess with USE_OPENGL, the user can select the renderer anyways before launch, the user can decide which renderer he wants to use based on his experience (artifacts, glitches, etc.).

If the computer which compiles the source doesn't have SSE2 then GCC will use optimizations for the specific CPU being used for compiling. This is what the -march=native does. This way I have about 5-6% better performance (compared to march=pentium3).

The build server or maybe any other people except me will have SSE2, every 64 bit CPU has SSE2. So, if somebody wants to release a 32 bit version for the game and using a decent CPU (64 bit at least) to compile, it will be compiled against Pentium 4 (with SSE2 of course).

Issue #128, opened 2019-08-19
credPlaySmk method starts with return


CommonLoon102:
It is in the mapedit (and now master too) branch. Is there any reason? Edit: yes. With new GCC, SMK loading crashes on Windows.

CommonLoon102:
When compiling it with GCC it is crashing for me too on Windows 7 64 bit.

CommonLoon102:
```
Thread 1 received signal SIGSEGV, Segmentation fault.
0x000007fefefb116b in msvcrt!memmove () from C:\Windows\system32\msvcrt.dll
(gdb) bt
#0  0x000007fefefb116b in msvcrt!memmove ()
   from C:\Windows\system32\msvcrt.dll
#1  0x00000000007ce787 in nblood!SDL_DYNAPI_entry ()
#2  0x000000000052b857 in ?? ()
#3  0x000000000022fc10 in ?? ()
Backtrace stopped: previous frame inner to this frame (corrupt stack?)

```

CommonLoon102:
```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
```

CommonLoon102:
Sometimes it crashes at `file.Open(fileName);` sometimes at `SmackerCommon::LogError("Can't open file " + fileName);` and sometimes it puts a medkit into the inventory with 0% when the demo is playing :D

CommonLoon102:
OK, I found the solution. Compile with GCC 7.3.0. NBlood 1.01 is used that too and it didn't crash either. I'm talking about the Windows release of course.

I followed these steps: https://wiki.eduke32.com/wiki/Building_EDuke32_on_Windows#MinGW-w64.2FGCC_.28Standalone.29

At installation, select:
Version: 7.3.0
Architecture: x86_64
Threads: win32
Exception: seh
Build revision: 0

```
mingw32-make clean
mingw32-make -j4
```

And it is not crashing anymore.

CommonLoon102:
GCC 4.8.5 also seems fine, Cheogh is also fine on Linux with 4.8.5, I think we should mention it in the readme.md that this is the recommended version for compiling NBlood (for EDuke32 the requirement is >=4.8).

CommonLoon102:
I think the issue is in somewhere in `kopen_internal`.

sirlemonhead:
Could you confirm on your end that this has been resolved with the latest commits relating to this issue?

Hendricks266:
[...]

PR #129, opened 2019-08-20
fix SMK playback

CommonLoon102:
Related issue: https://github.com/nukeykt/NBlood/issues/128
For some reason, this `return` was added when merging code from EDuke32 and it caused cinematics to not play anymore.


nukeykt:
SMK playback causes crashes on windows (it's fine on linux and i find nothing wrong with address sanitizer)

CommonLoon102:
On Windows 7 it doesn't crash. What OS do you use? Windows 8.1?

CommonLoon102:
OK, compiling it with GCC is causing the crash.

PR #130, opened 2019-08-20
Tiny Caleb scaling sync

CommonLoon102:
Sync mapedit with master regarding tiny Caleb scaling, so later we won't have huge tiny Calebs by mistake.


nukeykt:
No need for this change

CommonLoon102:
Won't be mapedit merged into master at some point? When it happens, will you remember that you have to change -4 to 64?

Issue #137, opened 2019-08-21
Cheogh is brown on Linux in mapedit branch


CommonLoon102:
Not just Cheogh, every large gray stone gargoyle is brown instead of gray.
On E1M7 it just flickers sometimes to brown, but usually it is gray.
On E4M2 it is always brown except when it shoots its projectile.
On E4M8 it is the same as on E4M2.

Only on Linux, on Windows they are gray :S
Master branch is OK in every case.
Both software and polymost renderers are affected.

Tested on 3 different Linux machines (2 physical, 1 virtual: 1 Lubuntu 18 32 bit, 1 Fedora 30 64 bit, 1 Ubuntu 18 64 bit).

CommonLoon102:
This commit is causing the issue: https://github.com/nukeykt/NBlood/commit/00b94cdc17ec0a4d15df5dbb041d16bf0597f1a0

CommonLoon102:
I compiled the game on Windows 7 with GCC and Cheogh is gray... So, it is not an issue with MSVC.

CommonLoon102:
It gets even better: if I compile with RELEASE=0 on Linux, then Cheogh is gray on Linux too. I start liking this bug :D So, anybody, any ideas how to move forward? :D

CommonLoon102:
If I get https://github.com/nukeykt/NBlood/commit/00b94cdc17ec0a4d15df5dbb041d16bf0597f1a0 and then cherry pick 94192284, a71e4456, ae845fd5, cace828b to make it compile on Linux then I can fix this issue by adding `COMMONFLAGS += -fno-ipa-icf` to Common.mak. Sadly, if I do the same for the latest version of mapedit branch, it won't fix the issue.

CommonLoon102:
The newest version needs `-fno-ipa-icf` instead.

CommonLoon102:
OK, bonus LULZ: I thought that I've now fixed the issue, but when I pressed escape and the main menu came in, Cheogh became brown again, but only when the menu is visible. TOP KEK: ONLY ON FEDORA :D On Ubuntu it works fine :D

CommonLoon102:
Possible solution: compile NBlood with GCC 4.8.5. Cheogh will be OK with that, also the cutscenes won't crash on Windows with GCC 4.8.5.

CommonLoon102:
[...]

PR #138, opened 2019-08-22
fix Cheogh being brown instead of gray on Linux

CommonLoon102:

At the end of the day, this is what I find the most proper fix.
If I only disable some compiler optimizations, it still won't be perfect on Fedora (see the related issue). Furthermore, Clang doesn't know these flags and wouldn't be able to compile.

If I set -O1 instead of -O2 then Cheogh's color will be OK both on Ubuntu and Fedora in every case.
But if I compile with Clang, only the RELEASE=0 mode produces correct colors for Cheogh. But Clang is crap, sometimes it even removes your code entirely[1].

Since I cannot test Mac, and I don't want to waste time with the crappy Clang, I decided to fix this only for GCC on Linux.

I decided to disable O2 optimizations and just leave O1 optimizations. First time disabling `ipa-icf` did the job, but when later the code changed I had to disable `ipa-cp`. Who knows what the future brings? It is better to stay on the safe side, the compiler wants to be too clever, but it just messes up Cheogh. This is even worse than a security flaw[1] :D

[1]: https://www.redhat.com/en/blog/security-flaws-caused-compiler-optimizations

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,7 @@ project.xcworkspace/
 
 .DS_Store
 ._*
+/movie
+/blud*.png
+/*.DEM
+/game*.sav
--- a/Common.mak
+++ b/Common.mak
@@ -454,7 +454,7 @@ else
     LTO := 1
 
     ifeq ($(PLATFORM),LINUX)
-        if (ifeq (0,$(CLANG))
+        ifeq (0,$(CLANG))
             # This needs to be 1, otherwise Cheogh's color would be brown instead of gray on Linux
             OPTLEVEL := 1
         endif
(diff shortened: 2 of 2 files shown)


CommonLoon102:
Or building with gcc 4.8.5 also fixes the issue...

nukeykt:
I believe this is either some undefined behavior or memory corruption. Have you tried to use address sanitizer?

CommonLoon102:
> 
> 
> I believe this is either some undefined behavior or memory corruption. Have you tried to use address sanitizer?

Now I tried that I commented out all the address sanitizer flags in Common.mak (and using -O2), but same result, Cheogh is brown with GCC 7.4.0. I also commented out the `$(F_JUMP_TABLES) $(F_NO_STACK_PROTECTOR)` part, but gives the same results.
Or what do you mean by

> Have you tried to use address sanitizer?

What should I try?

CommonLoon102:
I've closed this because I believe changing O2 to O1 is not the solution, the solution is to use GCC 4.8.5.

PR #139, opened 2019-08-23
fix SMK playback on Windows

CommonLoon102:
Related issue: https://github.com/nukeykt/NBlood/issues/128


nukeykt:
I don't think using old version is a "fix" here

CommonLoon102:
It is a fix, the new GCC is broken...

PR #141, opened 2019-08-23
fix incorrect palette usage on Linux

CommonLoon102:

Use optimization level 1 if building on Linux or BSD with GCC and GCC version is not 4.8.x.
Otherwise, some sprites will be rendered with wrong palettes (depending on where the sprite is on the map (floortype)). I tried GCC 5, 6, 7 and 9 but they are broken too, only 4.8.x works.

### DIFF
--- a/Common.mak
+++ b/Common.mak
@@ -453,7 +453,7 @@ else
     OPTLEVEL := 2
     LTO := 1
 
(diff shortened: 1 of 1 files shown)


CommonLoon102:
This is the correct solution: https://github.com/nukeykt/NBlood/pull/145

PR #145, opened 2019-08-25
extra can be -1 then pointer will point to random

CommonLoon102:
Took me 5 days :'(
Thank you C.


nukeykt:
haha, nice. Thanks for your effort

Clownacy:
I know this is a year late, but have you guys tried using address-sanitisation? I imagine it would have caught this issue, and it could find similar undetected ones too.

Issue #147, opened 2019-08-26
Caleb will stick to the sky


CommonLoon102:
When using the jump boots and jumping at a place where you will reach the top of the map, Caleb will stick to the "ceiling" for a second or two.
Issue is only with mapedit branch, master branch is OK.

This is also caused by: 00b94cdc like https://github.com/nukeykt/NBlood/issues/142
Maybe these two issues have something to do with each other? Both are height related.

Issue #142, opened 2019-08-24
Cheogh will retreat in mapedit branch


CommonLoon102:
In master branch he works fine, but in mapedit branch instead of coming to you at E1M7, he will fly top of the map the farthest from you and stays there, looks like he wants to get out of the map, lol.

Maybe it is related to the newly added z coordinates?

If I crouch on/near the wooden bridge he can come to me, but otherwise he just flies to the top of the map.

Everyday, regular, normal gargoyles are affected too, not just the big grays. They tend to fly high and draw circles and never ever coming back to the ground again.

CommonLoon102:
This commit is causing the issue: 00b94cdc
The same commit which screwed up the palettes on Linux.

PR #150, opened 2019-08-27
fix ceiling clipping

CommonLoon102:
I don't know what these numbers mean, but I reverted them and it fixed both of these issues:
https://github.com/nukeykt/NBlood/issues/147
https://github.com/nukeykt/NBlood/issues/142

I reverted the ceiling and floor hit related constants which were modified by 00b94cd.

I don't know what the value should be in 2d.cpp since that is new code and I also don't know what does it do:
https://github.com/nukeykt/NBlood/compare/mapedit...CommonLoon102:jumpbug?expand=1#diff-2dd42d3488848376205f3c0f165e3257R915


nukeykt:
Hell no. This was a vital change to support 16384 sprites/walls (compared to original 4096 sprites and 8192 walls limit)

PR #152, opened 2019-08-27
 fix ceiling clipping 

CommonLoon102:
This fixes both of these issues:
https://github.com/nukeykt/NBlood/issues/147
https://github.com/nukeykt/NBlood/issues/142

Actually, only one modification would be enough to fix the above two issues (this: https://github.com/nukeykt/NBlood/compare/mapedit...CommonLoon102:jumpbug?expand=1#diff-d1843bc245ca73331396a72859d13e74R4778), but I've changed back the values at the other places too, because who knows what else they would break. The problem is that the new (old) clipping mask caused to calculate the ceilingZ incorrectly if there were no ceiling tiles above, only the sky. It made ceilingZ to be 2^32.
https://github.com/nukeykt/NBlood/blob/66b23e34736fed93735ed6e5205341634f3fa4bf/source/blood/src/gameutil.cpp#L709

### DIFF
--- a/source/blood/src/actor.cpp
+++ b/source/blood/src/actor.cpp
@@ -4663,7 +4663,7 @@ void MoveDude(spritetype *pSprite)
             if (sector[nSector].lotag >= 612 && sector[nSector].lotag <= 617)
             {
                 short nSector2 = nSector;
-                if (pushmove_old(&pSprite->x, &pSprite->y, &pSprite->z, &nSector2, wd, tz, bz, CLIPMASK0) == -1)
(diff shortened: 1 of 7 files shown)


nukeykt:
These bits collide with eduke32's rotated wall cstat bit though, which i'd like to support. As workaround it's possible to mask these bits before calling engine's functions, i guess. I'll look into this and push proper fix.

CommonLoon102:
Fixed here: https://github.com/nukeykt/NBlood/commit/8554cc07238534d427507abd1e3fc2fbdfde366e

PR #158, opened 2019-09-08
Blue eyes for Cheogh when blasting

CommonLoon102:
In the original, Cheogh's palette is 5 (grayscale) because in the seq he is brown. Except when he is blasting his projectiles, then he is gray with blue lights. Because his palette is always 5, his eyes will be white instead of blue.
So this bug is in the original too, I just thought it is worth fixing.


CommonLoon102:
If I make this change disabled by default and create a console command for it, would it be OK for you? It doesn't affect gameplay.

nukeykt:
I don't think it's worth

PR #162, opened 2019-09-08
make Cheogh's eye color configurable by console setting

CommonLoon102:
I think this setting is more important as it could seem to be. There's already a religious debate between Blood players what color Cheogh's eyes should be. I personally know at least 5, including myself. Cheogh's eyes should be blue because the projectiles he fires are blue too, therefore it would make sense to have his eyes blue when firing. This setting could let players have it their own preferred way, so everybody could be happy. It doesn't affect gameplay, only a little cosmetic change. It also wouldn't give any advantage to players in multiplayer by having this turned on or off. There shouldn't be a new fork created just because of this. Please reconsider adding this.


CommonLoon102:
The proposed hardcoded fix can potentially break custom content (mods), if the setting is enabled. The only proper fix for this is to fix the QAV for Cheogh's blasting attack. It is only 1 frame, its pallette should be set to something else than 0, like 3.

More:
Cheogh consists of multiple SEQs. The SEQ when Cheogh is blasting is activated when his AI state changes to blasting. The SEQ when he is blasting is only 1  frame and using palette 0. And in the code in seq.cpp in UpdateSprite() (which is being called because new SEQ was spawned due to AI state change) the palette for the sprite is only updated if the next SEQ's first frame's palette is non-zero. This is "unfortunate". If I change anything, even with a cvar setting, it would potentially break mods and custom content. The only way to fix this is to modify the SEQ and set the palette to 3, as I see it looks the same when it is 0. So yeah, this particular bug is only possible to fix with a mod, if we don't want to potentially break already existing stuff.

PR #167, opened 2019-09-15
New HUD for CTF

CommonLoon102:
The original HUD is quite unusable for CTF games for the following reasons:

It prints "captured red flag" instead of "captured blue flag": this is now fixed.

You cannot know how many players are in the game, who they are and in which team.
You don't know anything about the state of the flags, are they stolen or in the base, etc. You don't even know if you have currently either of the flags at you.

I've removed the ugly BLUE and RED score indicators from the top corners and instead added back the player list display slots as it is in bloodbath mode. Instead of displaying the frags, it will display a blue F and/or a red F if the specific player has either or both flags.

I've added a blue and red flag icon to the right and the score is below them. If the enemy team has the flag it will have an exclamation mark on the flag with the enemy team's color. If one of the allied team's player has it then the exclamation mark will have the allied team's color.

If the flag is down (dropped) there will be a down arrow on the flag with the flag's color.

If you have the flag, the flag will be animated.


source/blood/src/player.cpp
```diff
@@ -1344,6 +1344,10 @@ void PickUp(PLAYER *pPlayer, spritetype *pSprite)
     {
         pickedUp = PickupItem(pPlayer, pSprite);
         if (pickedUp && customMsg == -1) sprintf(buffer, "Picked up %s", gItemText[nType - 100]);
+        if (147 == nType)
```
nukeykt: This should be implement in `PickupItem`

source/blood/src/blood.cpp
```diff
@@ -119,6 +119,9 @@ bool gPaused;
 bool gSaveGameActive;
 int gCacheMiss;
 
+bool gBlueFlagDropped = false;
```
nukeykt: I think this should reside in player.cpp

source/blood/src/view.cpp
```diff
@@ -1312,8 +1312,128 @@ void DrawPackItemInStatusBar2(PLAYER *pPlayer, int x, int y, int x2, int y2, int
...
+        char* name = gProfile[p].name;
+        if (gProfile[p].skill == 2)
+            sprintf(gTempStr, "%s", name);
+        else
+            sprintf(gTempStr, "%s [%d]", name, gProfile[p].skill);
+        Bstrupr(gTempStr);
+        viewDrawText(4, gTempStr, x + 4, y + 1, -128, 11 + col, 0, 0);
+
+        x += 74;
+        if (gPlayer[p].at90 & 2)
```
nukeykt: Use `viewDrawText` here

nukeykt:
Also, i think gRedFlagDropped and gBlueFlagDropped should be cleared on game restart

nukeykt:
Thanks

PR #173, opened 2019-09-18
Colorful view messages

CommonLoon102:
- Display trigger texts from INI files with gold color
- Display chat messages in multiplayer with dark blue


nukeykt:
Thanks

PR #174, opened 2019-09-18
Option to turn off 'picked up...' messages

CommonLoon102:
- Fix issue: messages on/off was not stored in config file
- New option: possibility to turn off 'picked up...' messages.  They don't really useful and they spam the message history hence hiding more important messages like trigger texts from INI files or player chat messages in multiplayer.


nukeykt:
I'm against of completely disabling it. I think adding priorities for messages would be better idea

CommonLoon102:
How would it work in practice? What is your idea?

nukeykt:
New message will be discarded if current messages have higher priority

PR #175, opened 2019-09-18
Show messages by priority

CommonLoon102:
Spamming 'Picked up...' messages won't hide more important INI map messages because messages are now displayed by respecting their priorities.

Also fixed the messages on/off setting which was not saved into the config file before.


source/blood/src/endgame.cpp
```diff
@@ -212,13 +212,14 @@ void CSecretMgr::Found(int nType)
         at8++;
     if (gGameOptions.nGameType == 0)
     {
+        int pal = VanillaMode() ? 0 : 2; // 2: light red
```
nukeykt: Revert this

source/blood/src/messages.cpp
```diff
@@ -306,118 +306,214 @@ void LevelWarpAndRecord(int nEpisode, int nLevel)
...
         Clear();
     }
-    else if (!at0 && state)
-        at0 = 1;
+    else if (!this->state && state)
+        this->state = 1;
 }
 
-void CGameMessageMgr::Add(const char *pText, char a2, const int pal)
+void CGameMessageMgr::Add(const char *pText, char a2, const int pal, const int priority)
```
nukeykt: Use enum for priority values

source/blood/src/player.cpp
```diff
@@ -1339,6 +1339,7 @@ void PickUp(PLAYER *pPlayer, spritetype *pSprite)
     char buffer[80];
     int nType = pSprite->type;
     char pickedUp = 0;
+    int messagePriority = 0;
```
nukeykt: Seems to be useless var to me?

PR #182, opened 2019-09-20
Only allow the host to start a new game

CommonLoon102:
To avoid chaos and trolling.


nukeykt:
I'm against of this idea

PR #184, opened 2019-09-21
Friendly fire related fixes

CommonLoon102:
- Fix voodoo doll ammo consumption and behavior: do not consume ammo on teammates via secondary attack; to not blind and make teammates drunk
- Landing sound on teammate: when jumping on top of teammate's head will make no sound instead of water splash sound
- Vector attacking teammates: blood won't be spawn, no pitchforking sound when attacking teammates

Teammates still can be wrapped by flames if you napalm them for example, but it is only their clothes burning, which is fine I think :)

Proxy bombs will blow up and push teammates but won't reduce their HP.


source/blood/src/teammate.h
```diff
@@ -0,0 +1,57 @@
...
+    return false;
+}
+
+inline bool IsTargetTeammate(spritetype *pSourceSprite, spritetype *pTargetSprite)
+{
+    if (!IsPlayerSprite(pSourceSprite))
+        return false;
+    PLAYER *pSourcePlayer = &gPlayer[pSourceSprite->type-kDudePlayer1];
+    return IsTargetTeammate(pSourcePlayer, pTargetSprite);
+}
```
nukeykt: Move these to player.h

PR #189, opened 2019-09-23
Make some hardcoded hotkeys configurable 

CommonLoon102:
The hotkeys for some menus and functionality which are triggered by the F keys (F1-F12) are hardcoded and cannot be changed. Now they can be changed. Speedrunners and save-scummers now can bind quick save and quick load to mouse thumb buttons.

I've also fixed the issue when you saved or loaded a game (not with quick save/load), your current selected inventory powerup item has activated automatically (if item use was set to enter).


nukeykt:
Imo only qucksave/load and third person view are worth adding. Your patch also uses all slots available for functional keys, eduke32 used them for controller related keys, so i was going to use them for future controller support as well.

nukeykt:
As of bbaabc1b4439f4bb9add1774956002bbd476a156 it is possible to bind quicksave/load to any key using `bind` console command

CommonLoon102:
You mean F1-F12 by functional keys?

nukeykt:
gamefunc_ stuff. Absolute maximum is 64 here, which you hit with this patch. Going further would involve changes to eduke32's mact library.

CommonLoon102:
You mean this line: `#define MAXGAMEBUTTONS      64`?
I don't see it used anywhere.

CommonLoon102:
> 
> 
> As of [bbaabc1](https://github.com/nukeykt/NBlood/commit/bbaabc1b4439f4bb9add1774956002bbd476a156) it is possible to bind quicksave/load to any key using `bind` console command

How can I bind it to mousebutton4 and 5?
Of course I could figure it out myself, but I don't think players would be able to do it. They won't even know about the console command. If they open settings.cfg then the mouse binds are not even there.
I think we would need to extend the max 64 bind limit if there is one at all, and add Quick Save, Quick Load and Screenshot to configurable gamefuncs.

CommonLoon102:
So, I tried to figure out how to bind mouse button 4 to quick save, I typed bind showkeys into the console and the game has crashed:

> Exception thrown at 0x6B101BCC (ucrtbased.dll) in nblood.exe: 0xC0000005: Access violation reading location 0x61637345.

```
void OSD_Printf(const char *fmt, ...)
{
    static char tmpstr[8192];
    va_list va;

    va_start(va, fmt);
    Bvsnprintf(tmpstr, sizeof(tmpstr), fmt, va); // Crashed here
    va_end(va);

    OSD_Puts(tmpstr);
}
```

CommonLoon102:
What a silly solution, storing every button state in one 64 bit integer, that's why we can't have more than 64 gamefunctions, lol.

PR #195, opened 2019-09-24
fix crash when typing bind showkeys into the console

CommonLoon102:
Title.


nukeykt:
I've merged eduke32 variant of fix in ea14f7a66dd96b9e6385c8ebec050ecfe65d86d0

Issue #194, opened 2019-09-24
Cultists don't respawn in multiplayer


CommonLoon102:
When monsters set to respawn. I think they (cultists) should respawn too.
Neither phantasms.

CommonLoon102:
Fixed: https://github.com/nukeykt/NBlood/commit/38ab27d90f2f8f04f29d5bb98c12ece275430816

PR #197, opened 2019-09-25
Fix that cultists don't respawn in multiplayer

CommonLoon102:
Related issue: https://github.com/nukeykt/NBlood/issues/194

Sprite type becomes some gib and therefore the if won't be true. The original sprite type before the dude died is stored in... in `zvel`. Makes sense, right?


nukeykt:
Apparently it is `zvel` field in disassembly as well. thanks for pointing out my mistake

Issue #196, opened 2019-09-25
Monster corpses disappear


CommonLoon102:
In case if the monsters are set to respawn in multiplayer.
For example dogs disappear after they die and their corpse will be at their respawn location. The corpse should be at their death location until they respawn. Maybe this is a bug in vanilla too.

PR #209, opened 2019-10-06
Fix disappearing dudes in coop with respawn

CommonLoon102:
Related issue: https://github.com/nukeykt/NBlood/issues/196

Every change is just comment or magic number removal.
The fix is one line, negating the condition of an if. I marked it.


source/blood/src/actor.cpp
```diff
@@ -6873,9 +6873,9 @@ bool actCheckRespawn(spritetype *pSprite)
             if (pXSprite->respawnPending == 1)
                 nRespawnTime = mulscale16(nRespawnTime, 0xa000);
             pSprite->owner = pSprite->statnum;
-            actPostSprite(pSprite->index, 8);
-            pSprite->flags |= 16;
-            if (pSprite->type >= kDudeBase && pSprite->type < kDudeMax)
+            actPostSprite(pSprite->index, kStatRespawn);
+            pSprite->flags |= kHitagRespawn;
+            if (!IsDudeSprite(pSprite))
```
CommonLoon102: This is the fix.

nukeykt:
I see you referenced statnum names from **leaked** alpha code, so i can't accept your pull request. Just leave a fix

CommonLoon102:
Actually, I've referenced them from BloodGDX.

nukeykt:
BloodGDX is definitely based on alpha code. Also you can't use any code from GDX because it's licensed under GPL3+ whereas my code is GPL2+

CommonLoon102:
Would it be OK if I change the macro names and comments?

CommonLoon102:
Maybe the macro names can stay, as these were already in:
#define kStatRespawn 8
#define kStatMarker 10
#define kStatFree 1024

The leaked source has the same names.

CommonLoon102:
Anyways, it doesn't even compile anymore :D
I'm closing this.

Issue #193, opened 2019-09-24
Coop crashes on The Lair of Shial


CommonLoon102:
> Assertion failed: pSprite->type >= kDudeBase && pSprite->type < kDudeMax in file \NBlood\source\blood\src\ai.cpp at line 298

It is in aiMoveForward.

Maybe monsters need to be set to respawn, I could reproduce it when it is set to respawn.
It happens when a little spider dies, and its picnum is 1784, its type will be 0. Its health is 137, I think it has something to do with monster respawning.

CommonLoon102:
> Assertion failed: pSprite->type >= kDudeBase && pSprite->type < kDudeMax in file \NBlood\source\blood\src\aispid.cpp at line 216


It is ThinkGoto.

CommonLoon102:
\NBlood\source\blood\src\actor.cpp(3723): Bad Dude Failed: initial=0 type=0 NORMAL

CommonLoon102:
> \NBlood\source\blood\src\actor.cpp(2811): Unexpected type in ConcussSprite(): Sprite: 3720  Type: 0  Stat: 6

CommonLoon102:
LOL, vanilla bug, I get the same error in DOSBOX.

PR #213, opened 2019-10-07
fix little spider respawn crash

CommonLoon102:
Related issue: https://github.com/nukeykt/NBlood/issues/193

Vanilla bug. Reproducible in DOSBox.
Steps to reproduce:
1. Start coop game on Lair of Shial, set monsters to respawn
2. Die near the mother spiders
3. Go back there and shoot the little spiders with flare gun secondary while they are near your corpse
4. The game will crash to DOS with this message: "src\actor.cpp(1801): Unexpected type in ConcussSprite(): Sprite: #### Type: 0 Stat: 6" where #### is the actual sprite number.

The same crash happens in NBlood, and very similar (probably the same) crash in BloodGDX too.

The little spiders' type will be 0, this is causing the issue. It only happens when monsters are set to respawn. The little spiders' picnum is a spider corpse at the moment when the crash happens.

This makes the game unbeatable in coop with respawning monsters.

I decided to just disable the respawn for little spiders, that fixes all other crashes due to type==0 (see related issue). The mother spiders are spawning new little spiders anyways, so it won't change a lot and makes the game possible to finish.


CommonLoon102:
I will use another approach, I will just comment out the error messages like NoOne did if it is a more accepted solution.

PR #218, opened 2019-10-15
fix little spider respawn crash

CommonLoon102:
Related issue: #193

Vanilla bug. Reproducible in DOSBox.
Steps to reproduce:

1. Start coop game on Lair of Shial, set monsters to respawn
2. Die near the mother spiders
3. Go back there and shoot the little spiders with flare gun secondary while they are near your corpse
4. The game will crash to DOS with this message: "src\actor.cpp(1801): Unexpected type in ConcussSprite(): Sprite: #### Type: 0 Stat: 6" where #### is the actual sprite number.

The same crash happens in NBlood, and very similar (probably the same) crash in BloodGDX too.

The little spiders' type will be 0, this is causing the issue. It only happens when monsters are set to respawn. The little spiders' picnum is a spider corpse at the moment when the crash happens.

This makes the game unbeatable in coop with respawning monsters.

Since @NoOneBlood already commented out the ConcussSprite error message, I decided that instead of disabling the respawn for little spiders, I will just comment out the other error messages too.
In the AI related methods I saw this "fix" in other files too, like in aigarg.cpp, so these should be fine too for the spiders and in general.


NoOneBlood:
It only happens if spiders was spawned, and not if they already was in map, right?

CommonLoon102:
> 
> 
> It only happens if spiders was spawned, and not if they already was in map, right?

Right. It only happens if monsters are set to respawn, this is only possible in multiplayer coop mode.

NoOneBlood:
This is not what i meant: i mean it only happens if "first incarnation" of spider already was placed on map by level designer and it was not spawned by boss or dudeSpawn marker? I'm testing it right now and it looks like this happens in second case

CommonLoon102:
I only reproduced this on Lair of Shial where boss spider spawns plus monsters set to respawn (coop). I haven't tried to reproduce it in other cases, but in single player there is no such issue so far.

NoOneBlood:
It looks like spiders can't get spawned (i see respawn effect on corpses, but nothing happens).
The thing is: there should not be this effect on corpses, it should be on init sprite coordinates.

It just looks they should not try to respawn, because they already was spawned via trigger and have no inittype and init coordinates.

In same map there is red spider which was placed in map by level designer and it works just fine

CommonLoon102:
So what would be your suggested fix?

NoOneBlood:
Haha, good question. I don't know yet :)

Issue #219, opened 2019-10-15
Coop: cultist respawns too early


CommonLoon102:
In coop if monsters set to respawn, cultists respawn too early after they died.
In DOS, it takes much more time (40 seconds instead of 10).

Related bug: https://github.com/nukeykt/NBlood/issues/212

After the last frame of the corpse, the cultist respawns, the last frame should be there for 20-30 seconds like in the original.

PR #220, opened 2019-10-15
fix cultist respawn time

CommonLoon102:
Related issue: https://github.com/nukeykt/NBlood/issues/219

It was a typo, wrong variable was used: `nMonsterSettings` instead of `nMonsterRespawnTime`.


nukeykt:
Thanks

PR #229, opened 2019-10-24
feature: voxel random angle offset

CommonLoon102:
Besides `rotate` now it is also possible to use `randomangoff` in def files for voxels. If `rotate` and `randomangoff` are both specified, only `rotate` will take effect.

Usually this is useful for barrels with dead bodies and maybe some other stuff. But I think the preferred way would be still maphacks.


nukeykt:
I think this feature is redundant. For official maps maps maphacks could be done and user maps could be modified to have properly aligned props

Issue #200, opened 2019-09-25
Randomize voxel angle


CommonLoon102:
There should be a parameter for voxel definitions to make some voxels have randomized angle offset. Fixing manually every barrel, candle, bottle, etc. via maphacks would be very time consuming and it is not really worth the time.

The other problem is that shells and bullet cases are always laying on the ground in the same direction which looks weird.

PR #230, opened 2019-10-25
randomize ejecting brass and shell angle on the ground

CommonLoon102:
Related issue: https://github.com/nukeykt/NBlood/issues/200

so if they are voxels they won't point to the same direction


nukeykt:
Add VanillaMode check because otherwise it'll break demos. Also why intmax? Angle is 0-2047 in build

PR #233, opened 2019-10-26
feature: option to show voxels instead of sprites for showing weapon

CommonLoon102:
above other players' head in multiplayer


source/blood/src/controls.cpp
```diff
@@ -291,7 +291,27 @@ void ctrlGetInput(void)
     if (BUTTON(gamefunc_Show_Opponents_Weapon))
     {
         CONTROL_ClearButton(gamefunc_Show_Opponents_Weapon);
-        gShowWeapon = !gShowWeapon;
+        switch (gShowWeapon)
```
nukeykt: Use (gShowWeapon+1)%3 here

nukeykt:
Tbh it looks ugly and is not consistent with player sprite (i.e smooth rotation looks odd compared to limited sprite rotation)

CommonLoon102:
> 
> 
> Tbh it looks ugly and is not consistent with player sprite (i.e smooth rotation looks odd compared to limited sprite rotation)

The sprites are ugly too, for example the voodoo doll is halfway into Caleb's head. The good thing is that you can choose which ugly version you want to use :)
Personally, I like the voxel version better, because I can see where exactly the other player is looking. On Discord I remember others were missing the voxel version too.

PR #234, opened 2019-10-27
feature: show player names

CommonLoon102:
feature: add option to show player names in multiplayer when looking at other players.

You cannot know who you are dealing with if there are more then 4 players in the game because there are only 4 trousers color in multiplayer.


source/blood/src/view.cpp
```diff
@@ -1257,6 +1257,25 @@ void viewDrawMapTitle(void)
     }
 }
 
+void viewDrawAimedPlayerName(void)
+{
+    if (!gShowPlayerNames || gMe->aim.dx == 0 || gMe->aim.dy == 0)
```
nukeykt: Should be (dx == 0 && dy == 0)

source/blood/src/view.cpp
```diff
@@ -1679,6 +1698,8 @@ void UpdateStatusBar(ClockTicks arg)
 
     if (gGameOptions.nGameType < 1) return;
 
+    viewDrawAimedPlayerName();
```
nukeykt: Move to viewDrawScreen alongside with viewDrawMapTitle

PR #235, opened 2019-10-27
maphacks: mdzoff for voxels WIP!

CommonLoon102:
Do not merge this yet.
For the z direction (height) it works both for classic and OpenGL renderer, but there is something wrong with x and y because they give different results between classic and OpenGL. Don't try to find logic in OpenGL code related to x and y, because I was just trying by trial and error :)

Do you have any idea how to make it work for x and y too? I'm trying to add this feature because the misaligned fire armor on E3E2, but that only requires the z direction which is OK with this change.

### DIFF
--- a/source/build/src/engine.cpp
+++ b/source/build/src/engine.cpp
@@ -6203,13 +6203,17 @@ draw_as_face_sprite:
             }
         }
 
+        x = tspr->x + spriteext[spritenum].offset.x;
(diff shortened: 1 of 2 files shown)


nukeykt:
i believe mdx/y/zoff are meant to offset rotation pivot for models (i.e used with roll and pitch) not the model itself

Issue #238, opened 2019-10-27
in Ubuntu 19.10 i get a segmentation fault without any other message just at the start


BielBdeLuna:
...compiling with RELEASE=0 to see if i can pinpoint where this fails

BielBdeLuna:
with RELEASE=0 I don't get any further messages besides the segmentation fault, compiling with DEBUGANYWAY=1

BielBdeLuna:
nope, I get the same segmentation fault without any other message, maybe RELEASE=0 and DEBUGANYWAY=1 only act when building the engine?

Jan200101:
having this same issue on Fedora 30 x86_64 with RELEASE=0 it compiles and runs just fine
checking with gdb I found this on the stack
```
Assertion failed: pItem != NULL in file source/blood/src/gamemenu.cpp at line 2596\n
```

BielBdeLuna:
ah yes, I'm also on x86_64.

CommonLoon102:
Same issue on Ubuntu 18, we can say the Linux port is broken.

BielBdeLuna:
it wasn't broken before on Ubuntu 19.4 as I could play it then, so something broke it in either in the change of version of Ubuntu or in the recent commits to the engine.

CommonLoon102:
The fix is already done, it will be merged very soon.

CommonLoon102:
It should be fine now.

BielBdeLuna:
fine! now IT lives agaaaiin!

PR #240, opened 2019-10-29
Fix the game on non-Windows

CommonLoon102:
Related issue: https://github.com/nukeykt/NBlood/issues/238


nukeykt:
Use ARRAY_SIZE (or whatever it called) macros instead

PR #242, opened 2019-11-02
add steam versions to search path

Jan200101:
the Linux build of NBlood had not added Fresh Supply or One Unit Whole Blood to the search path so I added them

### DIFF
--- a/source/blood/src/common.cpp
+++ b/source/blood/src/common.cpp
@@ -308,26 +308,15 @@ static int G_ReadRegistryValue(char const * const SubKey, char const * const Val
 static void G_AddSteamPaths(const char *basepath)
 {
     UNREFERENCED_PARAMETER(basepath);
-#if 0
     char buf[BMAX_PATH];
 
-    // PORT-TODO:
-    // Duke Nukem 3D: Megaton Edition (Steam)
-    Bsnprintf(buf, sizeof(buf), "%s/steamapps/common/Duke Nukem 3D/gameroot", basepath);
-    addsearchpath(buf);
-    Bsnprintf(buf, sizeof(buf), "%s/steamapps/common/Duke Nukem 3D/gameroot/addons/dc", basepath);
-    addsearchpath_user(buf, SEARCHPATH_REMOVE);
-    Bsnprintf(buf, sizeof(buf), "%s/steamapps/common/Duke Nukem 3D/gameroot/addons/nw", basepath);
-    addsearchpath_user(buf, SEARCHPATH_REMOVE);
-    Bsnprintf(buf, sizeof(buf), "%s/steamapps/common/Duke Nukem 3D/gameroot/addons/vacation", basepath);
+    // Blood: Fresh Supply
+    Bsnprintf(buf, sizeof(buf), "%s/steamapps/common/Blood", basepath);
     addsearchpath_user(buf, SEARCHPATH_REMOVE);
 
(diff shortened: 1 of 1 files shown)


source/blood/src/common.cpp
```diff
@@ -308,26 +308,15 @@ static int G_ReadRegistryValue(char const * const SubKey, char const * const Val
 static void G_AddSteamPaths(const char *basepath)
 {
     UNREFERENCED_PARAMETER(basepath);
```
Hendricks266: Should remove this line since this variable is used now.

source/blood/src/common.cpp
```diff
@@ -308,26 +308,15 @@ static int G_ReadRegistryValue(char const * const SubKey, char const * const Val
...
-    Bsnprintf(buf, sizeof(buf), "%s/steamapps/common/Duke Nukem 3D/gameroot", basepath);
-    addsearchpath(buf);
-    Bsnprintf(buf, sizeof(buf), "%s/steamapps/common/Duke Nukem 3D/gameroot/addons/dc", basepath);
-    addsearchpath_user(buf, SEARCHPATH_REMOVE);
-    Bsnprintf(buf, sizeof(buf), "%s/steamapps/common/Duke Nukem 3D/gameroot/addons/nw", basepath);
-    addsearchpath_user(buf, SEARCHPATH_REMOVE);
-    Bsnprintf(buf, sizeof(buf), "%s/steamapps/common/Duke Nukem 3D/gameroot/addons/vacation", basepath);
+    // Blood: Fresh Supply
+    Bsnprintf(buf, sizeof(buf), "%s/steamapps/common/Blood", basepath);
     addsearchpath_user(buf, SEARCHPATH_REMOVE);
```
Hendricks266: I believe this line should be changed to `addsearchpath(buf);`. Semantically, `SEARCHPATH_REMOVE` means the folder should no longer be considered for file reading after the RFF is loaded, which is never valid for Blood due to various files being loose in the directory, as well as CD music. This flag is used for Duke and SW to keep the virtual search path as clean as possible. If you don't observe any issue, it is probably because NBlood does not call the function to remove paths with this flag, but my point stands.

source/blood/src/common.cpp
```diff
@@ -308,26 +308,15 @@ static int G_ReadRegistryValue(char const * const SubKey, char const * const Val
...
 
-    // Duke Nukem 3D (3D Realms Anthology (Steam) / Kill-A-Ton Collection 2015)
-#if defined EDUKE32_OSX
-    Bsnprintf(buf, sizeof(buf), "%s/steamapps/common/Duke Nukem 3D/Duke Nukem 3D.app/drive_c/Program Files/Duke Nukem 3D", basepath);
+    // Blood: One Unit Whole Blood
+    Bsnprintf(buf, sizeof(buf), "%s/steamapps/common/One Unit Whole Blood", basepath);
     addsearchpath_user(buf, SEARCHPATH_REMOVE);
```
Hendricks266: Same as above.

source/blood/src/common.cpp
Jan200101: will do

source/blood/src/common.cpp
Jan200101: I was unaware of the difference of the two and just went with one of them
I will change this ASAP

nukeykt:
Merged, thanks
@Hendricks266 Thanks for reviewing

PR #244, opened 2019-11-07
search for gog installations

Jan200101:
the gog installer on Linux offers to put .desktop files in $HOME/.local/share/application and $HOME/desktop which contain the path to the game directory

this PR parses those desktop files and adds the found path to the search path

### DIFF
--- a/source/blood/src/common.cpp
+++ b/source/blood/src/common.cpp
@@ -318,6 +318,56 @@ static void G_AddSteamPaths(const char *basepath)
     addsearchpath(buf);
 }
 
+static void ParseDesktopFile(const char *desktop)
+{
+    buildvfs_fd fd = buildvfs_open_read(desktop);
+    int32_t size = buildvfs_length(fd);
+    char *dskbufstart, *dskbuf, *dskbufend, *dskpath;
+
+    if (size <= 0)
+        return;
+
+    dskbufstart = dskbuf = (char*)Xmalloc(size);
+    size = (int32_t)buildvfs_read(fd, dskbuf, size);
+    buildvfs_close(fd);
+    dskbufend = dskbuf + size;
+
+    char buf[BMAX_PATH];
+
+    while (dskbuf < dskbufend)
+    {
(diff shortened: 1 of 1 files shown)


Jan200101:
Ive found a slight fault in the code that causes it to seg fault when there is no newline at the end I'll resolve ASAP

source/blood/src/common.cpp
```diff
@@ -318,6 +318,56 @@ static void G_AddSteamPaths(const char *basepath)
...
+            break;
+        }
+        while (*dskbuf++ != '\n') {}
+    }
+
+    Xfree(dskbufstart);
+}
+
+
+static void G_AddGogPaths(const char *homepath)
```
Jan200101: Another way of adding the gog path would be, like AddSteamPaths, add the default directory

Hendricks266:
Hi Jan, thanks for making this PR. Please see r8286 (and r8287) in the mainline EDuke32 SVN for my take on the approach you've shown here. I've factored out much of the Steam and GOG path code into the engine to be shared by Duke 3D and SW, and when the NBlood repository is rebased I can make the same changes to Blood.

> Ive found a slight fault in the code that causes it to seg fault when there is no newline at the end I'll resolve ASAP

I believe my version does not have this problem, but if you could verify that for me I would appreciate it.

> Perhaps the function should accept a key to get the value from

This is probably only worth thinking more about if we ever find a use case for it.

> Should the data directory, which is used for the game content by dosbox, be changeable or toggable? For now there is no reason to because only One unit Whole Blood has a Linux gog release

For Blood's purposes, there is no need to add this functionality. In the future I intend to massively revamp our path detection to allow more flexibility of configuration, which is needed for the Redneck games and some distributions of SW's Twin Dragon addon.

> Another way of adding the gog path would be, like AddSteamPaths, adding the default directory

I had the same idea and implemented it in my patch.

Hendricks266:
Merged this functionality and confirmed there is no bug if a line with `Paths=` is not followed by a newline. Thanks again @Jan200101.

PR #247, opened 2019-11-22
Crash when exiting game on Windows 7

CommonLoon102:
fix crash on exit: game tries to write keyboard layout changed to xy log message after exit, but the log file is already null. fwrite cpp error: assert stream != nullptr


CommonLoon102:
obsolete, not reproducible anymore

Issue #201, opened 2019-09-25
Maphack x, y, z offset


CommonLoon102:
It would be nice to be able to offset voxels or sprites by x, y, z coordinates to fix some alignment issues, like on E3M2 the fire armor is halfway into the ground.

PR #257, opened 2019-12-31
Add possibility to move voxels/models via maphacks

CommonLoon102:
Related issue: https://github.com/nukeykt/NBlood/issues/201

This way the fire armor on E3M2 can be aligned above the ground. Just an example.
Maphacks are already available for BloodGDX and Fresh Supply. Those maphacks could be reused for NBlood if the md[xyz]off commands are replaced with pos[xyz]off commands.
I've done the changes based on @Hendricks266 guidance.

Works for eduke32 too, also in polymer renderer.

Quote from Hendricks:
> new tokens, new field
> don't change the definitions and uses of existing things
> assuming vec3_t offset is used for md[xyz]off, it may be helpful to rename it to pivot_offset or similar
> it may also be nice to add new aliases for the existing tokens reflecting that they are pivots
> there is another aspect to consider: how to write the rendering math when both tokens coexist
> fortunately in this case it seems simple and intuitive: apply pivots and rotations first, then shift positions
> or in other words, shift the pivot along with the model, and then have md[xyz]off shift the pivot further


Hendricks266:
If you prefix the new tokens with `md`, I'll merge it.

PR #262, opened 2020-01-04
Fix the SDL 1.2 and USE_OPENGL=0 build

BSzili:
Building with SDL 1.2 and without OpenGL was broken in Exhumed. Some of the fixes are inside Build and could apply to other games.

### DIFF
--- a/source/exhumed/src/movie.cpp
+++ b/source/exhumed/src/movie.cpp
@@ -285,14 +285,16 @@ void PlayMovie(const char* fileName)
                     bDoFade = DoFadeIn();
                 }
             }
+#ifdef USE_OPENGL
             else
             {
                 if (f >= 0)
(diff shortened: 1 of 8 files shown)


source/exhumed/src/movie.cpp
```diff
@@ -285,14 +285,16 @@ void PlayMovie(const char* fileName)
...
                 }
             }
+#ifdef USE_OPENGL
             else
             {
                 if (f >= 0)
                 {
                     fullscreen_tint_gl(0, 0, 0, f);
-                    f -= 8;
+                   f -= 8;
```
Hendricks266: Whitespace typo here

BSzili: Oops, fixed

Issue #268, opened 2020-01-10
Blood: Monolith and GTI intro logos


CommonLoon102:
I think we should add back to play Monotlith and GTI SMK files when starting Blood because a lot of people remember incorrectly and they expect them to play. In Blood version 1.21 they were disabled: https://blood-wiki.org/index.php?title=Bugs_in_Blood#Version_1.21_.28One_Unit:_Whole_Blood.29 (see first bug).

Because of this, people think that SMK playback doesn't work and that they couldn't set up correctly the `movie` folder. Users could see it is working by starting a new game on episode 1 but they usually don't do that they just expect the logos at the start.

We could add a -nointro parameter if there isn't one already.

edit: -quick is exactly for this already, so I didn't add -nointro

sirlemonhead:
Yes, I think this should be done. I'm one of the many who expect to see them (I don't think I ever updated/patched beyond the plasma back 'back in the day' - whatever version that was.

There is already a parameter, -quick, which should be used for this.

PR #269, opened 2020-01-10
Blood: Play Monolith and GTI intro logos

CommonLoon102:
Related issue: https://github.com/nukeykt/NBlood/issues/268

### DIFF
--- a/source/blood/src/blood.cpp
+++ b/source/blood/src/blood.cpp
@@ -94,7 +94,7 @@ unsigned int nMaxAlloc = 0x4000000;
 bool bCustomName = false;
 char bAddUserMap = false;
 bool bNoDemo = false;
-bool bQuickStart = true;
+bool bQuickStart = false;
 bool bNoAutoLoad = false;
 
 int gMusicPrevLoadedEpisode = -1;
--- a/source/blood/src/credits.cpp
+++ b/source/blood/src/credits.cpp
@@ -36,19 +36,19 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #include "sound.h"
 #include "view.h"
 
-bool Wait(int nTicks)
+char Wait(int nTicks)
 {
     totalclock = 0;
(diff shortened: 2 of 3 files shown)


CommonLoon102:
The flickering didn't occur to me, it was working fine. Then for some unknown reason it started to flicker when fading out. I restored this code from this PR where it didn't flicker for me, now this flickers too.

After the GTI.SMK has been played, the other in-game cinematics are broken: the lower half of the screen contains a still image about the last GTI.SMK frame.

I have no idea how to fix the GTI.SMK aspect ratio, even VLC gives back a different result compared to your screenshot.

I could solve your other bullets though, but not these three. I don't really care about these intro videos, they contain 0 value anyways. I'm closing this.

Hendricks266:
Can you push what you have?

sirlemonhead:
I can work on this if nobody else is interested.

CommonLoon102:
OK, I found out the flickering happens when fading out from the sprite logo only when the renderer is OpenGL, it works with software rendering. The problem(?) is in `gSetDacRange`.

CommonLoon102:
Using bool results in bad machine code. What could I say. C++ seems like a really mature language :D

CommonLoon102:
In software mode the other issue is also absent: when the lower part of the GTI logo is shown when playing cinematics later (e.g. starting new game on episode 1).

CommonLoon102:
Using the software renderer the title screen (gargolye with Blood and Plasma Pak) won't be visible if the SMKs were played.

CommonLoon102:
@sirlemonhead this is as far as I go.

Remaining issues:
- The GTI.SMK's aspect ratio is not good for some reason
- If the GTI.SMK played at start then game cinematics won't play back correctly (but only in case of OpenGL)
- The fade out effect doesn't work in OpenGL (happening when the SMK files are not available)

CommonLoon102:
Fixed in https://github.com/nukeykt/NBlood/commit/52ccfa68f60e65907e21cf0741f6fe0542279f08

PR #274, opened 2020-01-11
update gitignore

CommonLoon102:


.gitignore
```diff
@@ -80,3 +80,18 @@ project.xcworkspace/
 /*.INI
 /*.sav
 /*.cfg_orig
+*.DEM
+*.DEM_
```
Hendricks266: Can you prefix these with `/`?

PR #292, opened 2020-01-25
Fix builds without XMP

CommonLoon102:


NY00123:
Fix applied in f8b1ee37e89366bff9491afc83869304e8c6e221.

PR #298, opened 2020-02-02
Blood: Prevent accidental skipping of cinematics

### LINKED ISSUE #285: Final cutscene doesn't play
[This](https://youtu.be/-vIr-rTd5N0) cutscene doesn't play, despite killing Tchernobog, and the ending credits playing.

CommonLoon102:
Cinematics still can be skipped instantly by pressing Escape. But pressing any other button will have effect only after 2 seconds.

In the original game it is only possible to skip the video with the `Escape` key.

Fixes https://github.com/nukeykt/NBlood/issues/285

### DIFF
--- a/source/blood/src/credits.cpp
+++ b/source/blood/src/credits.cpp
@@ -249,7 +249,8 @@ void credPlaySmk(const char *_pzSMK, const char *_pzWAV, int nWav)
     ClockTicks nStartTime = totalclock;
 
     ctrlClearAllInput();
-    
+
+    CONSTEXPR kb_scancode escape = 0x1;
     int nFrame = 0;
     do
     {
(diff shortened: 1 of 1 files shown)


source/blood/src/credits.cpp
```diff
@@ -249,15 +249,16 @@ void credPlaySmk(const char *_pzSMK, const char *_pzWAV, int nWav)
...
+    CONSTEXPR kb_scancode escape = 0x1;
     int nFrame = 0;
     do
     {
         gameHandleEvents();
         if (scale((int)(totalclock-nStartTime), nFrameRate, kTicRate) < nFrame)
             continue;
 
-        if (ctrlCheckAllInput())
+        if (KB_KeyPressed(escape) || (nFrame > nFrameRate<<1 && ctrlCheckAllInput()))
```
nukeykt: I think one second would be enough

PR #300, opened 2020-02-02
Blood: guard gameplay changes by ifdefs

CommonLoon102:
Since these behaviors were not part of the original game, I'm not sure if they should be enabled by default even if vanilla mode is off. These are not obvious bugs, but they exist like this to prevent infighting between monsters, I guess.


NoOneBlood:
Already have same changes in upcoming commit

PR #305, opened 2020-02-07
Refactor of NOONE_EXTESNIONS:

NoOneBlood:
All functions and definitions was moved to nnexts.h and nnexts.cpp files
to minimize ifdefs and generally make code more clear


source/blood/src/ai.cpp
```diff
@@ -128,6 +116,20 @@ void aiNewState(spritetype *pSprite, XSPRITE *pXSprite, AISTATE *pAIState)
         pAIState->enterFunc(pSprite, pXSprite);
 }
 
+bool isUmmune(spritetype* pSprite, int dmgType, int minScale) {
```
CommonLoon102: Ummune?

source/blood/src/nnexts.h
```diff
@@ -0,0 +1,271 @@
...
+extern short gPhysSpritesList[kMaxSuperXSprites];
+extern short gProxySpritesCount;
+extern short gSightSpritesCount;
+extern short gPhysSpritesCount;
+
+// - FUNCTIONS ------------------------------------------------------------------
+bool nnExtEraseModernStuff(spritetype* pSprite, XSPRITE* pXSprite);
+void nnExtInitModernStuff(bool bSaveLoad);
+void nnExtProcessSuperSprites(void);
+bool nnExtIsUmmune(spritetype* pSprite, int dmgType, int minScale = 16);
```
CommonLoon102: Should be Immune?

NoOneBlood: nnExtIsImmune and isImmune acts a bit differently, i left isImmune for Cerberus fix only, and nnExtIsImmune could normally work only ifdefs is enabled

CommonLoon102: I know, but you wrote isUmmune and nnExtIsUmmune instead isImmune and nnExtIsImmune, so just a typo I guess (`I` is next to `U` on the keyboard).

NoOneBlood: oh, thanks :D

PR #309, opened 2020-02-10
Blood: Clear keys on new level, fixes #308

### LINKED ISSUE #308: Blood: in Coop, keys are not cleared on new level
When the keep keys multiplayer setting is ON, it will keep the keys even when the game transitions to a new level.

CommonLoon102:
Fixes https://github.com/nukeykt/NBlood/issues/308

Please squash on merge so the amend commits won't pollute the main branch history.

### DIFF
--- a/source/blood/src/blood.cpp
+++ b/source/blood/src/blood.cpp
@@ -713,7 +713,7 @@ void StartLevel(GAMEOPTIONS *gameOptions)
             }
             playerInit(i,0);
         }
-        playerStart(i);
+        playerStart(i, 1);
     }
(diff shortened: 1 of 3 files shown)


nukeykt:
Add vanilla mode guard

CommonLoon102:
This is the guard: `gGameOptions.bKeepKeysOnRespawn`.
It is a multiplayer setting and only affects co-op. Disabled by default.

CommonLoon102:
Are you sure vanilla mode guard is needed here? It seems unnecessary.

PR #313, opened 2020-02-12
Blood: Add User Map menu to episode select

### LINKED ISSUE #202: Custom map selection
Currently user needs to type in the name of the map, a list about available maps would be nicer.

CommonLoon102:
Currently the users are only able to play custom maps (if they don't have an .ini file) by entering commands into the console. Selecting the difficulty is not possible this way, the previously used difficulty will be used (by default Lightly Broiled). This is not really intuitive for users.

Fixes #202

### DIFF
--- a/source/blood/src/blood.cpp
+++ b/source/blood/src/blood.cpp
@@ -586,7 +586,7 @@ void StartLevel(GAMEOPTIONS *gameOptions)
         if (gEpisodeInfo[gGameOptions.nEpisode].cutALevel == gGameOptions.nLevel
             && gEpisodeInfo[gGameOptions.nEpisode].at8f08)
             gGameOptions.uGameFlags |= 4;
-        if ((gGameOptions.uGameFlags&4) && gDemo.at1 == 0)
+        if ((gGameOptions.uGameFlags&4) && gDemo.at1 == 0 && !Bstrlen(gGameOptions.szUserMap))
             levelPlayIntroScene(gGameOptions.nEpisode);
 
         ///////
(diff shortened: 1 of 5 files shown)


MexMeRcUrY:
I have downloaded this version 2 times to double check and i cannot see the custom map menu ....is there any special setup I have to make?

CommonLoon102:
> I have downloaded this version 2 times to double check and i cannot see the custom map menu ....is there any special setup I have to make?

You don't need to do any special setup. Just change the branch to `custmapmenu` in VS and click the play button.
Before clicking the play button in VS, make sure that you have set the view to `eduke32.sln` in Solution Explorer (at the top, there is the Switch Views button).  Then right click on the `nblood` project and click Set as StartUp Project. Then you can click Play (or hit F5).

MexMeRcUrY:
> > I have downloaded this version 2 times to double check and i cannot see the custom map menu ....is there any special setup I have to make?
> 
> You don't need to do any special setup. Just change the branch to `custmapmenu` in VS and click the play button.
> Before clicking the play button in VS, make sure that you have set the view to `eduke32.sln` in Solution Explorer (at the top, there is the Switch Views button). Then right click on the `nblood` project and click Set as StartUp Project. Then you can click Play (or hit F5).

Found it, Thank you for the advise!!
image

CommonLoon102:
Opened a new PR for this which is using the new `CGameMenuFileSelect` file selector component: https://github.com/nukeykt/NBlood/pull/340

PR #317, opened 2020-02-14
Blood: Stop infinitely playing burning fuse sound in multiplayer

### LINKED ISSUE #315: Dynamite & Aero Cans do not explode after death when you respawn fast, and sound gets stuck in map.
Tested and was replicated easily.

Take dynamite bundle and hold fire down, have 2nd player kill you.  If you do not respawn fast, and watch the wick burn down, the correct behavior happens, it explodes.  

Do the same thing as above, this time respawn while the wick is still burning.  The explosion does not ever happen, and you are left with the burning wick sound forever in the map.

This same issue happens with aero cans.  the noise level is much louder with this weapon.

CommonLoon102:
Fixes #315


source/blood/src/sfx.cpp
```diff
@@ -396,11 +396,14 @@ void sfxKill3DSound(spritetype *pSprite, int a2, int a3)
     }
 }
 
-void sfxKillAllSounds(void)
+void sfxKillAllSounds(spritetype *pSprite)
```
nukeykt: Make separate variant for sprites

CommonLoon102: I'm not sure if I should refactor the duplicated code into a separate method or it is not advised in this codebase because the method will be called inside a loop.

PR #322, opened 2020-02-21
Fix for duplicate INI files showing in launch screen when sending CMD Params

MexMeRcUrY:
Fix for duplicate INI files showing in screen when sending CMD Parameter nblood -ini <fileName>.ini and the INI already exists in the Nblood folder. It will keep the Nblood folder file name  in case its not spelled exactly the same in the input CMD -ini for (Linux Case Sensitivity Compatibility)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -95,3 +95,8 @@ project.xcworkspace/
 /platform/Windows/*.cfg_orig
 /platform/Windows/*.DEM
 /platform/Windows/*.DEM_
+/eduke32.exe.lastcodeanalysissucceeded
+/mapster32.exe.lastcodeanalysissucceeded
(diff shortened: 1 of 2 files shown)


clobber:
Just noting, that's a *ton* of whitespace and formatting changes they probably didn't bargain for.

MexMeRcUrY:
> Just noting, that's a _ton_ of whitespace and formatting changes they probably didn't bargain for.

i think i will need to redownload the original and make the changes again. not sure why all those spaces and stuff is there maybe because i ran the code-analysis.

MexMeRcUrY:
@clobber I have created the issue description and committed the fresh changes again into
 issue #323
https://github.com/nukeykt/NBlood/issues/323

Thnx for the headsup!

PR #325, opened 2020-02-22
Fix VBO setup for voxels

### LINKED ISSUE #320: macOS: Regression - On launch, another long delay with black screen before video shows up
macOS 10.14.6
MacBookPro11,5
Intel Iris Pro / AMD Radeon R9 M370X

This is the same problem again -- originally reported in https://github.com/nukeykt/NBlood/issues/48 and fixed by commit https://github.com/nukeykt/NBlood/commit/b76634d90e7b0fb1f003ba0a795c6da0a86756b7

`git bisect` (a MAJOR chore due to all the commits that broke compiling) traced the first bad commit to https://github.com/nukeykt/NBlood/commit/2cc4724291ea01c89a86dfea12fa208e2ac8d16a and I was able to determine it was the same chunk of code in `source/build/src/voxmodel.cpp`: https://github.com/nukeykt/NBlood/commit/2cc4724291ea01c89a86dfea12fa208e2ac8d16a#diff-8a1d4fae1ea09c73ddd7c85e9562f599R437-R446 (expand voxmodel.cpp to see the diff).

https://github.com/nukeykt/NBlood/blob/b468083f5d61dce379f718a5c1e0a2dd24e3865c/source/build/src/voxmodel.cpp#L437-L446

I reverted all these lines back to original chunk of code from commit https://github.com/nukeykt/NBlood/commit/b76634d90e7b0fb1f003ba0a795c6da0a86756b7 and that resolves the issue, but I don't know if that is meant to be the correct fix since the latest changes -- the intent is unknown.

Perhaps @alexey-lysiuk or anyone else can have a look?

Sidenote: I've noticed that keyboard input now appears to be broken. ~I haven't determined the commit but the in-game menus are unresponsive.~ See: #321

alexey-lysiuk:
This fixes #320

### DIFF
--- a/source/build/src/voxmodel.cpp
+++ b/source/build/src/voxmodel.cpp
@@ -435,14 +435,14 @@ void voxvboalloc(voxmodel_t *vm)
         index[(i*2+1)*3+2] = i*4+3;
     }
     GLint prevVBO = 0;
+    glGetIntegerv(GL_ELEMENT_ARRAY_BUFFER_BINDING, &prevVBO);
     glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, vm->vboindex);
     glBufferData(GL_ELEMENT_ARRAY_BUFFER, sizeof(GLuint) * 3 * 2 * vm->qcnt, index, GL_STATIC_DRAW);
-    glGetIntegerv(GL_ELEMENT_ARRAY_BUFFER_BINDING, &prevVBO);
     glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, prevVBO);
     prevVBO = 0;
+    glGetIntegerv(GL_ARRAY_BUFFER_BINDING, &prevVBO);
     glBindBuffer(GL_ARRAY_BUFFER, vm->vbo);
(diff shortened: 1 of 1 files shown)


source/build/src/voxmodel.cpp
```diff
@@ -435,14 +435,14 @@ void voxvboalloc(voxmodel_t *vm)
         index[(i*2+1)*3+2] = i*4+3;
     }
     GLint prevVBO = 0;
-    glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, vm->vboindex);
-    glBufferData(GL_ELEMENT_ARRAY_BUFFER, sizeof(GLuint) * 3 * 2 * vm->qcnt, index, GL_STATIC_DRAW);
     glGetIntegerv(GL_ELEMENT_ARRAY_BUFFER_BINDING, &prevVBO);
+	glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, vm->vboindex);
```
Hendricks266: There's a tab here

clobber:
@Hendricks266 As expected, this fixes the black screen at launch with Ion Fury as well, so this patch ought to make it upstream.

PR #331, opened 2020-03-01
- Event redirection support for modern types

NoOneBlood:
- New modern type "Condition" (WIP)
- Fix damage scale for custom dude
- Extend Impact and Vector flags so they can be used with dudes
- Impact condition can be triggered with missiles and explosions
- Updates for Teleport Target type (gModernMap)
- Updates for Effect Gen type
- Updates for Wind Gen type


NoOneBlood:
> 
> 
> The 'fallthrough' annotation in useEffectsGen, nnexts.cpp, line 1737 is not correct and generates a compiler warning. What is the intent here?

I was think it will fall to default case

PR #336, opened 2020-03-26
Blood: Fix duplicated INI files in startup window

### LINKED ISSUE #323: Blood: Duplicate INI files showing in Launch screen when sending CMD Parameters
NBlood issue with the duplicate  INI (case sensitivity fixed too for LINUX)  I fixed it by compare against the nBlood Folder INI file name and will keep the "correct" file name in your Nblood folder instead of using the one misspelled in the CMD line because of the case sensitive. 
Nblood INI Issue

MexMeRcUrY:
Blood: duplicated INI files in startup window fix #323


nukeykt:
Thanks

PR #340, opened 2020-04-01
Blood: Add User Map menu to episode select

### LINKED ISSUE #202: Custom map selection

CommonLoon102:
Currently the users are only able to play custom maps (if they don't have an .ini file) by entering commands into the console. Selecting the difficulty is not possible this way, the previously used difficulty will be used (by default Lightly Broiled). This is not really intuitive for users.

Fixes #202

### DIFF
--- a/source/blood/src/gamemenu.cpp
+++ b/source/blood/src/gamemenu.cpp
@@ -3007,18 +3007,7 @@ bool CGameMenuItemPassword::Event(CGameMenuEvent &event)
     return CGameMenuItem::Event(event);
 }
 
-CGameMenuFileSelect::CGameMenuFileSelect(const char* _pzText, int _nFont, int _x, int _y, int _nWidth, const char* _startdir, const char* _pattern, char* _destination)
-{
-    InitObject(_pzText, _nFont, _x, _y, _nWidth, _startdir, _pattern, _destination);
-}
-
 CGameMenuFileSelect::CGameMenuFileSelect(const char* _pzText, int _nFont, int _x, int _y, int _nWidth, const char* _startdir, const char* _pattern, char* _destination, void(*_onFileSelectedEventHandler)() )
-{
-    InitObject(_pzText, _nFont, _x, _y, _nWidth, _startdir, _pattern, _destination);
-    onFileSelectedEventHandler = _onFileSelectedEventHandler;
-}
-
-void CGameMenuFileSelect::InitObject(const char *_pzText, int _nFont, int _x, int _y, int _nWidth, const char *_startdir, const char *_pattern, char *_destination)
 {
     m_pzText = _pzText;
     m_nFont = _nFont;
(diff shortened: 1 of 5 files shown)


source/blood/src/gamemenu.cpp
```diff
@@ -3005,6 +3008,17 @@ bool CGameMenuItemPassword::Event(CGameMenuEvent &event)
 }
 
 CGameMenuFileSelect::CGameMenuFileSelect(const char* _pzText, int _nFont, int _x, int _y, int _nWidth, const char* _startdir, const char* _pattern, char* _destination)
+{
+    InitObject(_pzText, _nFont, _x, _y, _nWidth, _startdir, _pattern, _destination);
+}
+
+CGameMenuFileSelect::CGameMenuFileSelect(const char* _pzText, int _nFont, int _x, int _y, int _nWidth, const char* _startdir, const char* _pattern, char* _destination, void(*_onFileSelectedEventHandler)() )
```
nukeykt, line 3015: Remove other constructor, and set callback default value to nullptr

source/blood/src/gamemenu.cpp
```diff
@@ -3243,7 +3257,17 @@ bool CGameMenuFileSelect::Select(void)
     if (!findhigh[currentList])
         return false;
 
-    Bstrcat(destination, findhigh[currentList]->name);
+    char name[MAX_PATH];
```
nukeykt, line 3260: What's point of this change?

source/blood/src/menu.cpp
```diff
@@ -1559,6 +1571,15 @@ void SetWeaponSwitch(CGameMenuItemZCycle *pItem)
 
 extern bool gStartNewGame;
 
+void ShowDifficulties()
+{
+    gGameMenuMgr.Push(&menuDifficulty, 3);
```
nukeykt, line 1576: Seems to be too hackish

CommonLoon102: Yeah, like the whole game engine and everything.

source/blood/src/gamemenu.cpp
CommonLoon102: To fix the case when user goes back from difficulty selection to user map menu and then selecting a different map. The "file browser control" remembers the last selected destination so it is needed to update it correctly, not just concatenating the newly selected file. Without the change it would be `something.mapsomething.map` instead of just `something.map`.

source/blood/src/gamemenu.cpp
CommonLoon102: Pushed new commit for this.

source/blood/src/menu.cpp
CommonLoon102: Pushed new commit regarding this. Maybe I could reduce the hackishness a little.

PR #349, opened 2020-04-15
blood: fix inputSync when FPS drop low

### LINKED ISSUE #329: Blood: Dynamite bundle thrown too soon
If the FPS is capped below 30 (fps limit at 30 and fps offset is between -10 and -1 (inclusively)) then the game logic will have some weird things. For example, if I hold the primary fire key to throw a dynamite bundle, Caleb will throw it sooner than I release the primary fire key. The lower the FPS the sooner he throws.

MexMeRcUrY:
FIXES #329

### DIFF
--- a/source/blood/src/blood.cpp
+++ b/source/blood/src/blood.cpp
@@ -989,7 +989,8 @@ void ProcessFrame(void)
     char buffer[128];
     for (int i = connecthead; i >= 0; i = connectpoint2[i])
     {
-        gPlayer[i].input.buttonFlags = gFifoInput[gNetFifoTail&255][i].buttonFlags;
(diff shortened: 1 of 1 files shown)


nukeykt:
Does 48a4a0450e8441dfc188293244bd21c50557ce83 fix issue on your end?

CommonLoon102:
> 
> 
> Does [48a4a04](https://github.com/nukeykt/NBlood/commit/48a4a0450e8441dfc188293244bd21c50557ce83) fix issue on your end?

On my end it fixes the issue.

MexMeRcUrY:
yes it fixed my scenario too. thnx

PR #353, opened 2020-04-17
Blood: Displays the correct "usage" of the -j command

### LINKED ISSUE #352: Blood: Fresh supply search path not working
I was testing fresh supply files and i noticed that even that in the code the search path exists ("%s/steamapps/common/Blood") it does not load it, however "One unit whole blood" works fine.
image
image

-J Param as workaround does not work when One unit whole blood is found:
nblood.exe -j "D:\Program Files (x86)\Steam\steamapps\common\Blood"

image

MexMeRcUrY:
fixes #352


MexMeRcUrY:
i think this is not accurate after further testing in latest version

Issue #348, opened 2020-04-15
Blood: Cutscenes option not available for cooperative mode (Coop Cutscenes)


MexMeRcUrY:
I convinced a friend to play the game and it was his first time playing it and it was a shame a key part of the game intro "story" and "Ending movie" after you defeat the final boss and finish a complete episode with your friend is not playing even if the move folder is available in the NBlood folder.

MexMeRcUrY:
The Scope of this issue and progress:
-Cutscene not available in Coop mode - 100% Credits @CommonLoon102 
-integration with menu and input CMD. - 0%
-Cutscene not playing when using Fresh Supply files (OGV) - 80%

HandMeBacon:
It looks like there was once a large PR (#357) to add support for OGV files following the creation of this issue, but it was closed and this issue is still open two years later. Is this issue still being worked? If not, I have some free time this weekend I could use to try to create a new PR.

sirlemonhead:
Hi, No don't think anyone is working on this!

Hendricks266:
The most complicated part of the patch is setting up libtheora to build portably and cleanly. If you can figure that out, making use of it in the code should be straightforward.

martingalvan:
Hi, I arrived here from https://github.com/NBlood/NBlood/issues/967

> The most complicated part of the task is setting up libtheora to build portably and cleanly. If you can figure that out, making use of it in the code should be straightforward.

Would vcpkg help with that?

PR #357, opened 2020-04-30
Coop cutscenes

MexMeRcUrY:
enhancements discribed in #348


Hendricks266:
The sdltheora library is unnecessary. You only need libtheora. Look at EDuke32's use of libvpx. Ideally both formats should be supported by the functionality you introduce.

>  Files changed 457
> +137,282

:/

CommonLoon102:
Hi,

Wow, a lot of effort has been done! Nice!

Some things to consider:
- In blood.cpp, unnecessary whitespace changes at line 90, 626 and 1111.
- Use `Bstrcpy` instead of `strcpy`.
- Use `Bstrcat` instead of `strcat`.
- Use `Bstrlen` instead of `strlen`.
- Use `Bstrrchr` instead of `strrchr`.
- In `G_GetGameDir` there is some strange returns (`return("...")`).
- `G_GetGameDir` is a copy-pasta of `G_AddSearchPaths`, therefore it is duplicated code, it needs to be refactored.
- The video quality of the OGVs are terrible compared to the original. Therefore, if the user has both FS and OUWB installed, it should use the OUWB files instead. (I know that this is a draft and the order might have been changed for debugging reasons.)
- In `credits.cpp` change tabs back to spaces as it was before, because it is messing up the code comparison (there are other unnecessary whitespace changes too).
- Use `BMAX_PATH` instead of `MAX_PATH`.
- Use `BMAX_PATH` instead of `FILENAME_MAX`.
- For some reason when pressing escape when the cinametics are playing, the game will freeze for a few seconds and audio still plays for the video.
- In windowed mode when the credits or cinametics are playing there are two windows: an empty game window and a borderless window with the video playing in it. I can move the game window from behind the video window.
- In fullscreen mode the cinametics are stretched and not keeping the original aspect ratio if I'm playing in for example 640x480.
- In fullscreen mode the whole screen flickers every time when a video is about to play, I can see my other windows open in Windows for a very short period of time.
- `bool filefound` is an unused variable.
- In the menu it is somewhere called `cut-scenes`, somewhere else it is `cutscenes`.
- I don't think it is needed as a multiplayer setting, but currently the setting does nothing because it isn't synchronized between the players, it is not in gPacketStartGame.
- Need to test it on Linux at least, and test if it still compiles and runs with USE_OPENGL=0 and without SSE2, as I see libtheora adds some SSE2 requirements. It is not a problem if OGV video playback needs SSE2, but if somebody doesn't have the OGVs and doesn't have SSE2 either, the game should still work with the old videos

MexMeRcUrY:
Thank you very much for the detailed  feedback. ill work on everything you pointed out and inproove

Hendricks266:
>     * Use `Bstrcpy` instead of `strcpy`.
>     * Use `Bstrcat` instead of `strcat`.
>     * Use `Bstrlen` instead of `strlen`.
>     * Use `Bstrrchr` instead of `strrchr`.

Don't do these four. compat.h says: `DEPRECATED: Standard library prefixing`. We only use the prefix for a case like Bstrcasecmp where strcasecmp isn't always available, such as MSVC.

PR #362, opened 2020-05-17
Fix the USE_OPENGL=0 build

BSzili:
It's that time of the year again when I build NBlood without Polymost.


Hendricks266:
Thanks for the PR!

PR #367, opened 2020-05-26
continue searching for paths even after one was found

Jan200101:
This is another PR to address #352 but instead of changing the behavior of the `-j` flag the logic to find game data has been changed to continue searching even after one was found.


Jan200101:
thats a good idea but that would still mean that if someone deleted just the folders contents and keeps the folder intact it wouldn't continue searching

Jan200101:
Any feedback on this?
I still believe that simply checking the registry for the directory and ignoring anything else is a bad idea, perhaps the folder the registry entry points to should be checked first before throwing everything else out the window

Issue #395, opened 2020-08-24
PCExhumed videoSetGamma error. Pi3


Exarkuniv:
im able to get PCExhumed to compile with `make PCExhumed` and i have no errors.

im able to start the game and i can hear the game, but cant see anything. its just a black screen
this is what is in the log

im running Debian Stretch 4.14.98

```
Exhumed r11949-e7d1f7be2
Built Aug 24 2020 17:55:05, GCC 6.3.0, 32-bit 
Using /home/pi/NBlood/ for game data
Using /home/pi/.config/pcexhumed/ for game data
Initializing SDL 2.0.8
Using "x11" video driver
Searching for game data...
Using "STUFF.DAT" as main game data file.
Initializing OSD...
No game controllers found
Executing "settings.cfg"
Initialized 96.0M cache
Setting video mode 1920x1080 (8-bpp fullscreen)
Refresh rate: 60.00Hz
OpenGL information
 Broadcom VideoCore IV HW OpenGL ES 2.0
videoSetGamma(): That operation is not supported
videoSetGamma(): That operation is not supported
Initializing sound: SDL ALSA driver on bcm2835 ALSA, bcm2835 ALSA: 44.1 KHz stereo with 32 voices
videoSetGamma(): That operation is not supported
videoSetGamma(): That operation is not supported
videoSetGamma(): That operation is not supported
videoSetGamma(): That operation is not supported
videoSetGamma(): That operation is not supported
videoSetGamma(): That operation is not supported
Wrote SETUP.CFG
Wrote settings.cfg
```

im going to guess it just might not work

CommonLoon102:
Have you tried building it with `USE_OPENGL=0`?
More info: https://wiki.eduke32.com/wiki/Building_EDuke32_on_Linux#Build_options

Exarkuniv:
yes that was the first way i tryed and it errored out on me. 
it it will help i can try again and post the error

i was able to use `USE_OPENGL=0` when i built blood. and that starts just fine

Exarkuniv:
```source/exhumed/src/gun.cpp: In function ‘int CheckCloseRange(short int, int*, int*, int*, short int*)’:
source/exhumed/src/gun.cpp:309:67: error: ‘OSD_Printf’ was not declared in this scope
         OSD_Printf("%s %d: overflow\n", EDUKE32_FUNCTION, __LINE__);
                                                                   ^
Failed building obj/exhumed/gun.o from source/exhumed/src/gun.cpp!
[...]

PR #397, opened 2020-08-26
Rednukem: fix USE_OPENGL=0 build

CommonLoon102:
Fixes https://github.com/nukeykt/NBlood/issues/398
Reported here: https://github.com/nukeykt/NBlood/issues/395#issuecomment-681024334


Hendricks266:
Not sure about this. Ideally the DN64 functions such as AddExplosion could still be called, but without OpenGL their implementations wouldn't do anything (for now). This patch disables too much.

PR #410, opened 2020-10-14
Blood: Use kMaxXSectors instead of magic number

### LINKED ISSUE #408: Blood: Cannot open some doors
First demo goes out of sync when the first secret is not opening.
Also the grave with the flares cannot be opened and neither the next secret behind the bushes.
On E1M2 the fridge in the kitchen also cannot be opened.

Changing `kMaxXSectors` back to 512 solves the issue.

This change broke it:
https://github.com/nukeykt/NBlood/commit/4372228914392a7925eced5e229c5aa1fddba4a7#diff-bda7c0f2a3db0d760e27dc3c67866321f42e98fea58390bfaf1bf1a03848da77R27

CommonLoon102:
Fixes #408

### DIFF
--- a/source/blood/src/db.h
+++ b/source/blood/src/db.h
@@ -23,8 +23,8 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #pragma once
 
 #define kMaxXSprites 16384
-#define kMaxXWalls 16384
-#define kMaxXSectors 4096
+#define kMaxXWalls 512
+#define kMaxXSectors 512
 
 #ifdef NOONE_EXTENSIONS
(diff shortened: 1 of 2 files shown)


nukeykt:
Can you do this change without raising limits? I want check thoroughly before raising them.

CommonLoon102:
> Can you do this change without raising limits? I want to check thoroughly before raising them.

OK, done, I totally agree.
Actually, it should be `a3 < kMaxXSector`, so I will change accordingly. Good catch.

CommonLoon102:
One more remark from my side: the number 2048 was erroneous itself, as it would cause overindexing the `xsector` array which has only `kMaxXSector` elements, which was 512 at the time.

PR #428, opened 2020-12-05
PCExhumed: more big endian fixes

BSzili:
This PR fixes the BOOK.MOV and demo playback on big endian systems.


sirlemonhead:
Thank you! :)

PR #429, opened 2020-12-06
PCExhumed: Torch bouncing fix

BSzili:
This PR fixes the bouncy torches on level 5, and possibly many others. I ported this from PowerSlaveGDX.

### DIFF
--- a/source/exhumed/src/view.cpp
+++ b/source/exhumed/src/view.cpp
@@ -184,6 +184,14 @@ static void analyzesprites()
 
         pTSprite->pal = RemapPLU(pTSprite->pal);
 
+        // PowerSlaveGDX: Torch bouncing fix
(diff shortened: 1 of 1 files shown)


Hendricks266:
Do these torches bounce in the DOS version too?

BSzili:
Yes, it's a mapping mistake in some of the levels.

BSzili:
This small visual glitch was clearly unintentional and easy to fix, so I thought there's no harm in adding the fix to PCExhumed. I've been going through the levels, and found no side effects yet, other torch sprites with the correct cstat are not affected. 

before:
https://gyazo.com/a744f985262e4cd3f4b2fbeb8a129bb8
after:
https://gyazo.com/e55f6158ac12ad79e65e6d0daa02173d

PR #434, opened 2020-12-19
PCExhumed: Add Caps lock as the auto run toggle key

BSzili:
I noticed that auto run could only be changed via the cvar, so I added the Caps Lock key for it similar to other Build games.


sirlemonhead:
Nice one, thanks!

PR #440, opened 2021-01-16
Blood: fix big endian build

jonathangray:
Fixes: 622ff258e ("Blood: Rename some SEQ variables to their proper human readable names.")

Patch from @jasperla


Hendricks266:
Thanks!

PR #441, opened 2021-01-16
PCExhumed: Misc code cleanup

BSzili:
There are no functional changes (hopefully), I only did some code cleanup to fix warning, and make it a bit more readable at places.


sirlemonhead:
Thanks! 👍

PR #450, opened 2021-02-07
PCExhumed: Don't try to open the CD track every frame

BSzili:
If the CD track file didn't exist the game loop kept calling playCDtrack every iteration without checking the return value. Now it only tries to play the track once if the file doesn't exist.


sirlemonhead:
Thanks again! :)

PR #455, opened 2021-02-18
EWitchaven: minor code cleanup

fhomolka:
Replaced some magic numbers, replaced some of the integers with booleans, removed other games from searchpathtypes


sirlemonhead:
Thanks!

PR #460, opened 2021-03-07
PCExhumed: Demo recording big endian fix

BSzili:
I added byteswaps to menu_DemoGameSave where necessary, and also fixed a mistake I previously made in WritePlaybackInputs.


sirlemonhead:
Thanks!

PR #462, opened 2021-03-09
Blood: Fix next/previous weapon switching in case of lighter

CommonLoon102:
Fix the issue when Caleb puts away the lighter and then pulls out again if the player is using the next/previous weapon button. No need to put away the lighter if spray/TNT is equipped and then TNT/spray will be equipped.


nukeykt:
Thanks

PR #469, opened 2021-04-18
Blood: Fix key 7 related crash

CommonLoon102:
Key 7 is only used in custom maps. It was causing a crash when a monster dropped it (pointer pointed to random).

This is a fix from Raze by coelckers:
https://github.com/coelckers/Raze/commit/d2c9b5979d7de4d6f5a756664184168a98af829b


coelckers:
Just a quick note: It doesn't crash in NBlood, but it does access undefined memory (picanm[-1] and sprite[-1]).

I also have a request regarding the commit message. Please read this: https://forum.zdoom.org/viewtopic.php?f=49&t=71985
In light of the problems mentioned in that discussion and some people getting really annoyed about it, can you please remove the @ from the message?

sirlemonhead:
Thanks! :)

PR #471, opened 2021-04-28
Move SDL_Quit() call to engineUnInit() to prevent segfault on KMSDRM and Wayland SDL2 backends.

vanfanel:
There's a well know and longstanding issue with MESA that causes a segfault if SDL_Quit() is called from atexit().
This affects the KMSDRM and WAYLAND SDL2 backends: when the game is run on KMSDRM or Wayland, it segfaults and locks the whole system on quit if SDL_Quit() is called from atexit().

This simple PR is a workaround for that.


CommonLoon102:
This can be closed now.

PR #473, opened 2021-05-05
PCExhumed: Add a crosshair

BSzili:
I noticed a rather funky looking crosshair sprite among the game's tiles, so I added it to the game. The position could probably be tweaked a bit, but it's more or less where it should be.


sirlemonhead:
That's great, thanks! :)

PR #474, opened 2021-05-09
PCExhumed: improved precache

BSzili:
I made some additions to the precache:
- animated tiles
- sequence pictures
- numbers for the countdown on the last level

This pretty much eliminates the tile loading during normal gameplay as long as everything fits in the cache.


sirlemonhead:
Thanks!

PR #485, opened 2021-06-18
NBlood: fix building without NOONE_EXTENSIONS

BSzili:
After the latest additions viewAddEffect failed to build without NOONE_EXTENSIONS. I also added ifdef guards at a couple of other places.


sirlemonhead:
Thanks!

PR #492, opened 2021-07-24
NBlood: fix a resource leak in netDeinitialize

BSzili:
The gNetENetInit check in netDeinitialize was redundant, and it could cause some resource leak (the socket not getting closed) when the network game was aborted early on the "waiting for players" screen.


sirlemonhead:
Thanks!

PR #493, opened 2021-07-25
NBlood: fix some UI elements affected by the screen size

BSzili:
A few UI elements shrunk when the screen size as reduced, for example the images next to the save/load list, or the background of the "LEVEL STATS" text. This is a vanilla bug, but it's just a visual glitch on the UI, so I added the missing cstat flags where I saw this issue.


nukeykt:
Thanks

PR #502, opened 2021-08-17
Add gamepad support to RedNukem

sirlemonhead:
Add gamepad support to RedNukem (Duke3D, Redneck Rampages and Duke Nukem 64).

The only issue I'm aware of is that Next/Previous weapon changing isn't working as expected in the Redneck Rampage games. Some weapons get skipped (e.g. crowbar in Redneck Rampage). This needs to be looked at but isn't working expected even using Next/Previous from the keyboard so can be looked at separately.


sirlemonhead:
Committed each piece of this manually to avoid the 'Merge branch 'master' commits.

PR #505, opened 2021-08-21
[Blood] Renamed untitled spider functions

carnivoroussociety:
This PR renames the various untitled functions/variables for AI spider class.


nukeykt:
Thanks

PR #514, opened 2021-09-13
[Blood] Added negative number support for DrawStatNumber()

carnivoroussociety:
This adds color palette toned minus characters to DrawStatNumber() so negative numbers can be represented. Additionally this required updating `nblood.pk3` with the new png files as well as updating the `blood_widescreen.def` file.

In-game examples:
example
ctf

Additionally, I've included a number of small fixes from my own personal fork - if this is out of scope for this PR I'm happy to close and split each fix into their own PR.


Hendricks266:
For `earL0` and `earR0`, I'll accept a commit that makes them `static`, but I don't agree with renaming them. Giving variables a subscript 0 suffix, pronounced "nought", is standard notation for a thing's starting value or origin point.

carnivoroussociety:
I wasn't entirely aware of that, in that case I'll revert that naming commit, thanks!

PR #521, opened 2021-10-11
Blood: added an enum for weapon types

BSzili:
I added human readable names for the different weapon types, and applied them to weapon.cpp. There are probably a couple more places where they could be used, but 99% of it should be covered.


NY00123:
Been going through a few PRs, and manually applied this one in 7e27d153fc61be73c7f2aa1d364deaf48f09314d. Seen no change that matters in output binary or disassembly of it. Looks good, thanks!

Issue #487, opened 2021-07-18
NBlood - Keypad [.] does not work when typing IP in multiplayer


SP1LL:
The keypad numbers work but not the decimal/period

Also I would recommend it being able to save the last IP entered so you don't have to erase localhost manually every time

Hendricks266:
Can you confirm the `.` character can be added from the keypad (the key between KP0 and KPEnter)?

PR #528, opened 2021-11-06
Blood: fix delete event not being accepted in edit and bitmap fields

ExtraDrowsy:
This is meant to address issue #487. 

For the edit and bitmap menu field types, it looks like we don't accept delete events that are emitted (the numpad [.] technically doubles as delete key as well so it emits a delete event).

This affects menu items that are edit/text fields or ones that are linked to bitmaps (when typing in a save file name). Note that this fix appears to work fine with the actual non-numpad delete key as well (it does nothing rather than breaking).

### DIFF
--- a/source/blood/src/gamemenu.cpp
+++ b/source/blood/src/gamemenu.cpp
@@ -1907,6 +1907,7 @@ bool CGameMenuItemZEdit::Event(CGameMenuEvent &event)
         return false;
     case kMenuEventKey:
     case kMenuEventSpace:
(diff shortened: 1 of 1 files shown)


NY00123:
Been going through a few PRs, and manually applied this one in 483a7ac5b79c43bb0b2e96e82a51d88c5fde4fc7. Seems to work for entering IP or a saved game's name, thanks!

ExtraDrowsy:
@NY00123 I appreciate it, thanks!

PR #534, opened 2021-12-18
- fixed two issues in condCheckMixed:

coelckers:
* check for ceilingstat used ceilingshade. Use of floorstat in the same block was correct.
* some sector check functions accessing walls.

I asked NoOne about these, he confirmed for both that it was not correct.


nukeykt:
thx

PR #538, opened 2022-01-02
fixed sprite interpolation on rotating sectors.

coelckers:
The ->Drag case only updated the interpolation coordinates after changing them, which is too late.
I left the original behavior in for vanilla mode.

Corresponding bug report: https://github.com/coelckers/Raze/issues/613


coelckers:
Sorry, branch was not clean

PR #539, opened 2022-01-02
- Blood: fix interpolation of sprites being dragged along on rotating…

coelckers:
The ->Drag case only updated the interpolation coordinates after changing them, which is too late.
I left the original behavior in for vanilla mode.

Corresponding bug report: https://github.com/coelckers/Raze/issues/613


NY00123:
Been going through a few PRs, and manually applied this one in d83d090693ab52ac8946de6cb3c6b4644c32eea3. Seems to resolve the problem in E1M4, thanks!

Issue #494, opened 2021-07-26
NBlood: The delirium effect is too fast


BSzili:
Compared to the DOS version the delirium shake is too fast. The duration is the same, but it seems to do two full cycles instead of just one.

CommonLoon102:
What about the software renderer? It should be the same in that case. In polymost it is using an OpenGL solution, basically completely re-implemented from scratch. It just looks similar.

BSzili:
The software renderer and Polymost are the same in this regard, they "play" the delirium effect twice. This is probably a timing issue that affects both renderers.

sirlemonhead:
Should be fixed in PR #541 if you'd like to retest? Thanks!

BSzili:
Absolutely, now the delirium looks like it does in the DOS version.

sirlemonhead:
Perfect, thanks everyone!

PR #541, opened 2022-01-16
Blood: fix delirium being twice as fast

ExtraDrowsy:
Meant to address issue #494. I compared the original DOS executable and confirmed it is indeed faster. This change does not change the overall duration of the effect (still about ~8 seconds), just the speed of it.

### DIFF
--- a/source/blood/src/view.cpp
+++ b/source/blood/src/view.cpp
@@ -3233,7 +3233,7 @@ void viewUpdateDelirium(void)
     if ((powerCount = powerupCheck(gView, kPwUpDeliriumShroom)) != 0)
     {
         int tilt1 = 170, tilt2 = 170, pitch = 20;
-        int timer = (int)gFrameClock*4;
+        int timer = (int)gFrameClock << 1;
         if (powerCount < 512)
         {
             int powerScale = (powerCount<<16) / 512;


source/blood/src/view.cpp
```diff
@@ -3233,7 +3233,7 @@ void viewUpdateDelirium(void)
     if ((powerCount = powerupCheck(gView, kPwUpDeliriumShroom)) != 0)
     {
         int tilt1 = 170, tilt2 = 170, pitch = 20;
-        int timer = (int)gFrameClock*4;
+        int timer = (int)gFrameClock << 1;
         if (powerCount < 512)
```
ExtraDrowsy, line 3237: Just multiplying by 2 instead of 4 but by bit shifting one position, even if this is unnecessary optimization hehe.

ExtraDrowsy: I believe there is also a possibility that something else is modifying the clock upstream before it gets to this code, though I didn't see anything that looked wrong (I might be wrong). Also a possibility the clock class (`ClockTicks`) is newer so maybe it doesn't behave exactly like the original game's clock did? Nonetheless I don't believe anybody has reported any other issues with things that the clock is used for, also the code for `ClockTicks` used in EDuke is identical (hasn't had anything tweaked or fixed), so it seems like it's more prudent to make the change locally here (at where the clock value ends up being used), rather than somewhere upstream or in the clock code itself...

sirlemonhead:
Thanks!

ExtraDrowsy:
Hey thanks!

PR #542, opened 2022-01-18
BLOOD: add option to share keys in multiplayer

zZeck:
Selecting the shared option means any player picking up a key gives it to all players. Zandronum has an option like this for Doom ([SV_ShareKeys](https://wiki.zandronum.com/DMFlags#ZaDMFlags)). Makes casual coop more fun when familiarity with the game varies in the group.

Thanks for NBlood. I've spent many hours playing the game solo and multiplayer with it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,6 +91,8 @@ xcuserdata/
 project.xcworkspace/
 *.dSYM/
 
+.vscode
+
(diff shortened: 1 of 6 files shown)


sirlemonhead:
Thanks!

PR #543, opened 2022-01-25
BLOOD: clear gPost actions on actInit. Fixes some crashes.

zZeck:
An easy crash to trigger: Start a multiplayer game, and then end the game as host, while players are connected. netPlayerQuit sets up a post action, to delete sprite gPlayer[0]. This will end up happening after the demo is loaded that immediately tries to play, leading to a crash with an assertion failure for nSprite < kMaxSprites && sprite[nSprite].statnum < kMaxStatus or similar.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,6 +91,8 @@ xcuserdata/
 project.xcworkspace/
 *.dSYM/
 
+.vscode
+
 .DS_Store
 ._*
 /source/voidwrap/sdk/
--- a/source/blood/src/actor.cpp
+++ b/source/blood/src/actor.cpp
@@ -2385,8 +2385,8 @@ int nDudeToGibClient2 = seqRegisterClient(DudeToGibCallback2);
 int gPostCount = 0;
 
 struct POSTPONE {
-    short at0;
-    short at2;
+    short nSprite;
+    short nStatus;
 };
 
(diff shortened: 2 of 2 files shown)


carnivoroussociety:
This introduces a bug where loading state will always reset gPost array.

zZeck:
> This introduces a bug where loading state will always reset gPost array.

Thanks for the review.
Updated:

    void actInit(bool bSaveLoad) {
        if(!bSaveLoad)
            memset(gPost,0,sizeof(gPost));

Since ActorLoadSave::Load calls with true, right after setting gPost

    Read(gPost, sizeof(gPost));
    actInit(true);

That's the issue you were referring to yes?

carnivoroussociety:
Yes, but there's another issue. Instead of clearing gPost, gPostCount should be reset back to 0 for new level. That's because the level could end while gPostCount is above zero, and on next tick it'll try to process sprite 0 with status 0. You may want to do this instead.
```c
    if (!bSaveLoad) { // reset postpone count on new level
        gPostCount = 0;
    }
```
This will ensure whatever values in gPost will not be processed (array is never initialized to begin with).

zZeck:
>This will ensures hatever values in gPost will not be processed (array is never initialized to begin with).

Done. Thanks again.

sirlemonhead:
Thanks!

PR #544, opened 2022-01-25
BLOOD: fix host end game causing clients to ShutDown

zZeck:
Currently, when the host ends a game, all the clients ShutDown, which is frustrating, and seems like a bug. (I thought they were crashing, but they just make it to the end of app_main). This PR sets the gQuitRequest to 2, so ProcessFrame's gRestartGame = gQuitRequest == 2 will set gRestartGame true.

Note, that without the [gPost fix](https://github.com/nukeykt/NBlood/pull/543) the client has post action sprite deletion problems, leading to a fairly quick crash during demo playback.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,6 +91,8 @@ xcuserdata/
 project.xcworkspace/
 *.dSYM/
 
+.vscode
+
 .DS_Store
 ._*
 /source/voidwrap/sdk/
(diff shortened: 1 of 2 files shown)


zZeck:
Updated. Fixed issue where Quit on host no longer Quit, but restarted instead.

sirlemonhead:
Thanks!

Issue #191, opened 2019-09-23
Out of sync between Linux and Windows


CommonLoon102:
It goes out of sync for the host right after somebody moves. The game runs well though, but the out of sync text will be on the screen for the host.

Hendricks266:
Moving discussion here from #545.

Hendricks266:
@zZeck wrote:

> [MinGW by default](https://stackoverflow.com/questions/24015852/struct-packing-and-alignment-with-mingw) was using [ms_struct](https://gcc.gnu.org/onlinedocs/gcc-10.3.0/gcc/x86-Type-Attributes.html#x86-Type-Attributes) style packing. This made XSPRITE a have different structure in memory, and different size. Checksums don't match as a result. In network.cpp CalcGameChecksum this code
> 
> ```
>     sum = 0;
>     length = sizeof(XSPRITE)/4;
>     while (length--)
>     {
>         sum += *pBuffer++;
>     }
> ```
> 
> would give a different sum.

PR #545 specified `-mno-ms-bitfields` to GCC as a fix.

@nukeykt wrote:
> disabling check for xsprite checksum altogether might be a better idea. ENet code is robust enough I think

NY00123:
> Fixed as of [1af3d49](https://github.com/nukeykt/NBlood/commit/1af3d499f5ebad602689e78cd86f8b9b4813f2e6)

That seemed to fix syncing of XSPRITE data, but there are other fields in the PLAYER struct turning out to not being calculated in the same manner.

There should be a commit resolving this. I think that this issue may remain open for a short while (e.g., a day or two), at least until after a few more tests are behind us.

NY00123:
Hopefully fixed for good after applying 1c72fbc436aaeab33b3f8ecfba6a5674bca01fc6 (along with 1af3d499f5ebad602689e78cd86f8b9b4813f2e6).

PR #545, opened 2022-01-27
BLOOD: fix linux windows multiplayer desync

zZeck:
Fixes [Out of sync between Linux and Windows #191](https://github.com/nukeykt/NBlood/issues/191).

[MinGW by default](https://stackoverflow.com/questions/24015852/struct-packing-and-alignment-with-mingw) was using [ms_struct](https://gcc.gnu.org/onlinedocs/gcc-10.3.0/gcc/x86-Type-Attributes.html#x86-Type-Attributes) style packing. This made XSPRITE a have different structure in memory, and different size. Checksums don't match as a result. In network.cpp CalcGameChecksum this code

        sum = 0;
        length = sizeof(XSPRITE)/4;
        while (length--)
        {
            sum += *pBuffer++;
        }

would give a different sum.

This should fix GCC and Clang builds at least. Not sure what should be done about MSVC build on windows connecting to GCC build on linux.

I suppose the other option is to ALWAYS force ms style packing. Unless there's a way to change MSVC's packing behavior.


nukeykt:
disabling check for xsprite checksum altogether might be a better idea. ENet code is robust enough I think

NY00123:
I think that I recently got this reproduced (i.e., an out of sync message during what still looked like a functional game with no sync issue), albeit while using NotBlood.

So, maybe a fix can be applied? I currently think about two options:
1. Disable the xsprite checksum check as suggested by @nukeykt.
2. Calculate the checksum in a cross-platform manner, similar in idea to struct serialization.

Hendricks266:
Let's move discussion to #191.

PR #546, opened 2022-01-29
BLOOD: fix last user map to not override menu map selected for new game

zZeck:
After a user map has been chosen the map that appears selected on the multiplayer config menu is never used again, even when the multiplayer menu is re-loaded for a New Game.  This makes it impossible to play the built in maps without restarting the executable.

The problem relates to StartNetGame and zUserMapName

    Bstrncpy(gPacketStartGame.userMapName, zUserMapName, sizeof(gPacketStartGame.userMapName));
    gPacketStartGame.userMap = gPacketStartGame.userMapName[0] != 0;

Nothing ever resets zUserMapName to an empty string. This means after any user map is chosen, gPacketStartGame.userMapName is always populated, and gPacketStartGame.userMap is always true.

I added a custom CGameMenuItem for the multiplayer menu title text with an Event method that clears zUserMapName on kMenuEventInit, which is only sent when the menu is re-loaded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,6 +91,8 @@ xcuserdata/
 project.xcworkspace/
 *.dSYM/
 
+.vscode
+
 .DS_Store
 ._*
 /source/voidwrap/sdk/
--- a/source/blood/src/gamemenu.cpp
+++ b/source/blood/src/gamemenu.cpp
@@ -565,6 +565,19 @@ CGameMenuItemTitle::CGameMenuItemTitle(const char *a1, int a2, int a3, int a4, i
     bEnable = 0;
 }
 
+CGameMenuItemMultiplayerTitle::CGameMenuItemMultiplayerTitle(const char *a1, int a2, int a3, int a4, int a5) : CGameMenuItemTitle(a1, a2, a3, a4, a5)
+{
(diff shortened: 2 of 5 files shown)


carnivoroussociety:
I'm glad someone else found this issue. The only issue I see is that once you select a user level you're forced to reopen the menu in order to clear `zUserMapName`. Perhaps make it so `SetupNetLevels()` also clears `zUserMapName`, and add a new function to `itemNetStart3` so it clears `zUserMapName` as well.

zZeck:
SetupNetLevels so menu episode change would clear it, and itemNetStart3 so menu level change would clear it. I'll get that fixed tomorrow. Good catch!

I also need to fix the map and changelevel console commands (they have this same problem), but that's a rather different code path, so that'll be a different PR whenever I get around to that. Probably going to work on some other stuff before it.

carnivoroussociety:
Yeah, supporting changelevel for MP sounds like a pain but I'm glad someone is looking into it. Good job!

zZeck:
Alright, now changing the episode or level on the menu will also clear the user map.

sirlemonhead:
Thanks!

PR #547, opened 2022-01-29
Blood: Added negative number support for CTF HUD

carnivoroussociety:
This adds color palette toned minus characters to DrawStatNumber() so negative numbers can be represented (only large fonts are supported such as the CTF score display). Additionally this required updating `nblood.pk3` with the new png files as well as updating the `blood_widescreen.def` file.

In-game examples:
example
ctf


source/blood/src/view.cpp
```diff
@@ -1138,10 +1138,46 @@ void DrawStatNumber(const char *pFormat, int nNumber, int nTile, int x, int y, i
...
+        if (tempbuf[i] == ' ')
+            continue;
+        if (tempbuf[i] == '-')
+        {
+            switch (nTile)
+            {
+            default:
+                numTile = kSBarNegative;
+                break;
+            case 2240:
```
zZeck, line 1154: What is going on with case 2240? [GDX indicates ](https://gitlab.com/m210/BloodGDX/-/blob/master/core/src/ru/m210projects/Blood/Types/Hud/HudRenderer.java#L5)
```java
	public final int kBigRed = 2190;
	public final int kBigBlue = 2240;
```
I do not see a 2190 case?

source/blood/src/view.cpp
```diff
@@ -1831,6 +1860,10 @@ void viewPrecacheTiles(void)
         tilePrecacheTile(kSBarNumberArmor2 + i, 0);
         tilePrecacheTile(kSBarNumberArmor3 + i, 0);
     }
+    for (int i = 0; i < 6; i++)
```
zZeck, line 1863: 6 could be a constant HUD_NUMBER_COLORS_COUNT or something? The other init loops in here all use magic numbers I believe, but it is confusing IMO.

zZeck:
negatives

Seems to work. Easy way to test is suiciding in team mode. Going negative then positive works fine.

Is there any way to end up with negative health or armor?

zZeck:
Hard coded for testing
```C++
sprintf(tempbuf, pFormat, -1000);
```
testingExtremeCase

Not sure how much of the following is relevant since I forced the number:

Border pixel around negative sign is too bold compared to the numbers themselves?
The negative sign for health seems to be slightly too dark of red?
Small text needs small negative sign (see armor and ammo)
Medium text needs medium negative sign (see medpack)
Negative sign for armor numbers is wrong color
Negative sign for ammo center and lower right is wrong color

source/blood/src/view.cpp
carnivoroussociety: 2190 would be covered with the default case. And I don't think nblood has those default defines.

source/blood/src/view.cpp
carnivoroussociety: You're right, I'll add that thank you.

carnivoroussociety:
I'll look at redoing the assets to match the number thickness then, along with matching the color tone for the default 2190 font set. Thanks for the feedback.

carnivoroussociety:
I've made the following changes:
* Adjusted the border pixel for the font so it isn't as contrasting
* Better matched red color tone for health font type
* Skip negative tiles for small/unsupported fonts (this PR is only addressing the large/CTF fonts)

As for defining the various magic numbers, I feel that should be addresses in a separate PR as I feel it is outside the scope of this PR.

example

zZeck:
Nice! Thanks again for fixing the bugged score display. Couldn't really play teams.

sirlemonhead:
Thanks!

PR #548, opened 2022-01-29
Blood: Fix shadow offset when rendering through room over room

carnivoroussociety:
This PR fixes a calculation offset error introduced by commit https://github.com/nukeykt/NBlood/commit/df4778e5130df5c11dc1d6bbd387b13498e9b356 when rendering a shadow above a ror floor. Instead of applying the secondary room position offset it'll use the height of the secondary room - this will correctly draw the Z offset required by the shadow.


zZeck:
What's a good map and location to go to see this in effect?

carnivoroussociety:
E2M1, if you pause at the right moment you're able to see the shadow correctly rendered from above the fire key room.

sirlemonhead:
Cheers!

PR #549, opened 2022-01-29
Blood: Rename various functions

carnivoroussociety:
This rename sub_5BCF4() to ResetCheats(), and sub_746D4() to fxSpawnBloodPod(). It also renames CDemo's at59ef to nDemosFound.

### DIFF
--- a/source/blood/src/fx.cpp
+++ b/source/blood/src/fx.cpp
@@ -289,7 +289,7 @@ void fxSpawnBlood(spritetype *pSprite, int a2)
     }
 }
 
-void sub_746D4(spritetype *pSprite, int a2)
+void fxSpawnPodBlood(spritetype *pSprite, int a2)
 {
     UNREFERENCED_PARAMETER(a2);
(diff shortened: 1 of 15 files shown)


source/blood/src/fx.cpp
```diff
@@ -289,7 +289,7 @@ void fxSpawnBlood(spritetype *pSprite, int a2)
     }
 }
 
-void sub_746D4(spritetype *pSprite, int a2)
+void fxSpawnBloodPod(spritetype *pSprite, int a2)
```
zZeck, line 292: what's a blood pod?

carnivoroussociety: Well fxSpawnBloodPod() operates the same as fxSpawnBlood() but for green/red pod enemies.

zZeck: OOOOOH. Maybe fxSpawnPodBlood() would be better? Could just be me.

carnivoroussociety: I agree, fixed.

sirlemonhead:
Thanks!

PR #550, opened 2022-01-29
Blood: Scale viewBurnTime() flames to window width

carnivoroussociety:
This PR will scale burn time flames to output window width instead of fixed 4:3.

Example:
16x9 ratio

### DIFF
--- a/source/blood/src/view.cpp
+++ b/source/blood/src/view.cpp
@@ -3047,14 +3047,17 @@ void viewBurnTime(int gScale)
         {
             nScale = scale(nScale, gScale, 600);
         }
-        int nStat = burnTable[i].nStat;
-        if (r_usenewaspect) // stretch to custom aspect ratio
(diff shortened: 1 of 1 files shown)


zZeck:
Same animation frame, at 1440x900
Before
original
After
scaled

It looks like the stretch is cutting off a few pixels on the sides.
Is the pixel stretching fine? Or would tiling it be better (if feasible)?

carnivoroussociety:
I went with stretched the sprites as the previous commit just used even spacing to spread out the individual flames themselves - and the cutting off point is accurate to how it renders for 4:3. Maybe there is a way to dynamically generate the spacing like the blood dripping QAV for menus but I haven't figured out a way to do that.

carnivoroussociety:
I've reverted the scaling tweak so it'll only reposition each flame according to the window output.

sirlemonhead:
Thanks!

PR #551, opened 2022-01-29
Blood: Fix voxel maphacks desyncing when load save

carnivoroussociety:
This PR adds spriteext array to file save format, and fixes the maphacks getting desynced when loading saves (e.g: blood voxel pack). For example, loading a save to a different level will carry over the current level maphacks.


nukeykt:
BYTEVERSION should be updated because this PR changes save format

carnivoroussociety:
BYTEVERSION has been +1 https://github.com/nukeykt/NBlood/blob/b172052003e7e154df65b5f848dc3cc0f16d1d93/source/blood/src/common_game.h#L43

sirlemonhead:
Thanks!

PR #552, opened 2022-01-29
Blood: Restore custom fov after rendering crystal ball powerup

carnivoroussociety:
This PR will fix the fov bug where using the crystal ball powerup will force the default fov to be rendered for the player's current view. After rendering the crystal ball powerup window it'll reset to the current fov.


sirlemonhead:
Thanks!

PR #553, opened 2022-01-29
Blood: Fix crash when trying to play next demo

carnivoroussociety:
This PR fixes a crash when attempting to play the next demo when only 1 demo has been found. This occurs when manually playing back a demo for a mod, in which after the demo ends it'll attempt to play the next demo (which does not exist). Instead this will check if there is more than 1 demo available to play.


source/blood/src/demo.cpp
```diff
@@ -406,7 +406,7 @@ void CDemo::Playback(void)
                 if (v4 >= atf.nInputCount)
                 {
                     ready2send = 0;
-                    if (at59ef != 1)
+                    if (at59ef > 1)
```
zZeck, line 409: rename at59ef to demoCount ?

carnivoroussociety: That's done in another RP actually. https://github.com/nukeykt/NBlood/pull/549

zZeck: 👍

sirlemonhead: Would you mind renaming the variable to match #549 if you don't mind please? I'm not particularly familiar with Git so I'd probably end up messing something up if I tried to change this PR myself (if I even can)! 

Thanks for all your help so far :)

carnivoroussociety: No problem, hopefully this is all that needed to be changed to resolve the merge conflict. Thanks!

sirlemonhead: Hmm no still get a conflict. I could just manually push the change as a separate commit?

carnivoroussociety: I'm not sure myself, I've merged with master on my end so yeah you may have to do that.

sirlemonhead:
I've manually committed this change in commit a34ace2 . Thanks! :)

PR #554, opened 2022-01-29
Blood: Don't process already free'd fx sprites

carnivoroussociety:
This PR fixes a rare edge case where it'll try to process sprites that are set to be free'd. This can occur when triggering too many blood sprites on a single tick, which this will cause the game to crash on next tick.


sirlemonhead:
Thanks!

PR #555, opened 2022-01-29
Blood: Check if sound is enabled for sndStartWavDisk()

carnivoroussociety:
This PR will fix a bug where sound will play for sndStartWavDisk() even if the player has sound muted.


NY00123:
Looks like a side-effect of the original game not appearing to have an in-game sound toggle.
Added in c1cca9343456c467935ec8f86320a3e237da4951, thanks!

PR #556, opened 2022-01-29
Blood: Fix green/fire pod behavior

carnivoroussociety:
This fixes two issues with the current implementation for green/fire pods.

Fire pod:
Fixed issue where it always had to hit a floor surface in order to explode. 1.21 DOS has this set to always explode on any collision hit, which I have correctly matched.

Green pod:
1.21 would only do a range check for whatever it hit (0-4095), and introduced a bug where hitting a wall would then use that wall ID as the sprite index for damage. This has been preserved for vanilla mode flag, and for non-vanilla mode (NBlood) it will check if the hit object is a sprite before applying sprite damage.


sirlemonhead:
Thanks!

PR #558, opened 2022-01-29
Blood: Set max audio voices to 255

carnivoroussociety:
This fixes a bug introduced by commit https://github.com/nukeykt/NBlood/commit/6902fc73eaecfa2cd049970d8b69f8808aee897b. Instead of using 256 max voices, it will be set to 255. This is because voices are clamped between 0-255 by MV_MAXVOICES define.

### DIFF
--- a/source/blood/src/menu.cpp
+++ b/source/blood/src/menu.cpp
@@ -640,7 +640,7 @@ CGameMenuItemZBool itemOptionsSound3DToggle("3D AUDIO:", 3, 66, 80, 180, false, 
 CGameMenuItemSlider itemOptionsSoundSoundVolume("SOUND VOLUME:", 3, 66, 90, 180, &FXVolume, 0, 256, 48, UpdateSoundVolume, -1, -1, kMenuSliderPercent);
 CGameMenuItemSlider itemOptionsSoundMusicVolume("MUSIC VOLUME:", 3, 66, 100, 180, &MusicVolume, 0, 256, 48, UpdateMusicVolume, -1, -1, kMenuSliderPercent);
 CGameMenuItemZCycle itemOptionsSoundSampleRate("SAMPLE RATE:", 3, 66, 110, 180, 0, UpdateSoundRate, pzSoundRateStrings, 3, 0);
-CGameMenuItemSlider itemOptionsSoundNumVoices("VOICES:", 3, 66, 120, 180, NumVoices, 16, 256, 16, UpdateNumVoices, -1, -1, kMenuSliderValue);
+CGameMenuItemSlider itemOptionsSoundNumVoices("VOICES:", 3, 66, 120, 180, NumVoices, 16, 255, 16, UpdateNumVoices, -1, -1, kMenuSliderValue);
 CGameMenuItemZBool itemOptionsSoundCDToggle("REDBOOK AUDIO:", 3, 66, 130, 180, false, UpdateCDToggle, NULL, NULL);
 CGameMenuItemZCycle itemOptionsSoundMusicDevice("MIDI DRIVER:", 3, 66, 140, 180, 0, UpdateMusicDevice, pzMusicDeviceStrings, ARRAY_SIZE(pzMusicDeviceStrings), 0);
 CGameMenuItemChain itemOptionsSoundSF2Bank("SF2 BANK", 3, 66, 150, 180, 0, &menuOptionsSoundSF2, 0, NULL, 0);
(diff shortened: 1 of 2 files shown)


zZeck:
Seems like MV_MAXVOICES should be 255 then if 255 is the max voices?
Then
```
    Bassert(Voices < MV_MAXVOICES);
```
should be
```
    Bassert(Voices <= MV_MAXVOICES);
```

And the code here can then be more meaningfully written as
```
    CGameMenuItemSlider itemOptionsSoundNumVoices("VOICES:", 3, 66, 120, 180, NumVoices, 16, MV_MAXVOICES, 16, UpdateNumVoices, -1, -1, kMenuSliderValue);
```
```
    { "snd_numvoices", "the number of concurrent sounds", (void *)&NumVoices, CVAR_INT, 1, MV_MAXVOICES },
```

carnivoroussociety:
I'm not sure myself, and I don't feel comfortable making commits that'll adjust the audiolib functions themselves in fear of breaking something else entirely. I'll still go in and use that define as the max anyways, thanks!

zZeck:
> I'm not sure myself, and I don't feel comfortable making commits that'll adjust the audiolib functions themselves in fear of breaking something else entirely. I'll still go in and use that define as the max anyways, thanks!
Makes sense.
Leaving MV_MAXVOICE at 256 then and in your code doing MV_MAXVOICES -1 would be a good alternative to the raw 255 I think

source/blood/src/menu.cpp
```diff
@@ -328,9 +328,9 @@ CGameMenuItemChain itemNetStart12("START GAME", 1, 66, 175, 280, 0, 0, -1, Start
 CGameMenuItemText itemLoadingText("LOADING...", 1, 160, 100, 1);
 
 CGameMenuItemTitle itemSoundsTitle("SOUNDS", 1, 160, 20, 2038);
-CGameMenuItemSlider itemSoundsMusic("MUSIC:", 3, 40, 60, 180, MusicVolume, 0, 256, 48, SetMusicVol, -1, -1);
-CGameMenuItemSlider itemSoundsSound("SOUND:", 3, 40, 70, 180, FXVolume, 0, 256, 48, SetSoundVol, -1, -1);
-CGameMenuItemSlider itemSoundsCDAudio("CD AUDIO:", 3, 40, 80, 180, CDVolume, 0, 256, 48, SetCDVol, -1, -1);
+CGameMenuItemSlider itemSoundsMusic("MUSIC:", 3, 40, 60, 180, MusicVolume, 0, MV_MAXTOTALVOLUME, 48, SetMusicVol, -1, -1);
```
zZeck, line 331: MV_MAXTOTALVOLUME is defined as 255 so these are good

source/blood/src/menu.cpp
```diff
@@ -640,7 +640,7 @@ CGameMenuItemZBool itemOptionsSound3DToggle("3D AUDIO:", 3, 66, 80, 180, false,
 CGameMenuItemSlider itemOptionsSoundSoundVolume("SOUND VOLUME:", 3, 66, 90, 180, &FXVolume, 0, 256, 48, UpdateSoundVolume, -1, -1, kMenuSliderPercent);
 CGameMenuItemSlider itemOptionsSoundMusicVolume("MUSIC VOLUME:", 3, 66, 100, 180, &MusicVolume, 0, 256, 48, UpdateMusicVolume, -1, -1, kMenuSliderPercent);
 CGameMenuItemZCycle itemOptionsSoundSampleRate("SAMPLE RATE:", 3, 66, 110, 180, 0, UpdateSoundRate, pzSoundRateStrings, 3, 0);
-CGameMenuItemSlider itemOptionsSoundNumVoices("VOICES:", 3, 66, 120, 180, NumVoices, 16, 256, 16, UpdateNumVoices, -1, -1, kMenuSliderValue);
+CGameMenuItemSlider itemOptionsSoundNumVoices("VOICES:", 3, 66, 120, 180, NumVoices, 16, MV_MAXVOICES, 16, UpdateNumVoices, -1, -1, kMenuSliderValue);
```
zZeck: MV_MAXVOICES is defined as 256 though, so you need to -1 it I believe

carnivoroussociety: Yep, I just realized haha.

carnivoroussociety:
Unfortunately it looks like both MV defines are only defined in the private header `_multivoc.h` so I'll default back to raw values. All the other games do the same so I imagine this is the 'proper' solution.

sirlemonhead:
Thanks!

PR #559, opened 2022-01-29
Blood: Document FXDATA struct variables

carnivoroussociety:
Clean up variables used for fxSpawn and set fxSpawn()'s duration override argument to 0 by default.


sirlemonhead:
Thanks!

PR #560, opened 2022-01-29
Blood: Update gSaveGameOptions() on save game

carnivoroussociety:
This PR will fix the incorrect difficulty setting listed when saving using game menu. SaveGame() will use UpdateSavedInfo() the same way as QuickSaveGame(), and update the gSaveGameOptions array with current session settings.


sirlemonhead:
Thanks!

PR #562, opened 2022-01-30
BLOOD: save host, join, and multiplayer settings on use. Load on start.

zZeck:
Makes it much more convenient when the game saves last used settings. That way don't have to keep entering friend's IP address, and preferred game settings. The save load code is similar in style to what is used in loadsave.cpp for saving games.

A step toward loading config of choice from file potentially.


carnivoroussociety:
Is it possible for this to use the OSD system for saving/loading instead of managing the cfg files manually within menu.cpp?

zZeck:
> Is it possible for this to use the OSD system for saving/loading instead of managing the cfg files manually within menu.cpp?

What do you have in mind? I have not looked at the OSD code much.

carnivoroussociety:
So both zNetAddressBuffer/zNetPortBuffer are stored as a CVAR, which are automatically save/load handled within `nblood_cvars.cfg`

Something like this
`{ "net_address","sets network address used for multiplayer menu", (void *)zNetAddressBuffer, CVAR_STRING|CVAR_FUNCPTR, 0, 16 },`
Which works without needing to manually manage cfg files. I've done something like this for my own fork https://github.com/carnivoroussociety/NotBlood/commit/abc1d2602acd6451430da12d30a614d909e49ded.

carnivoroussociety:
Saving host/port feature has now been merged with OSD functionality.

PR #566, opened 2022-02-04
Blood: Don't show keys on HUD for bloodbath/teams mode

carnivoroussociety:
This PR will add a check to key icon drawing for gViewSize 1/2. Because bloodbath/teams mode equips all key on spawn, there is no reason to show the keys on the HUD. So if the player is using gViewSize 1/2, and is currently in a bloodbath/teams match, it will not draw currently held keys. gViewSize 3 and above (full size HUD) will still draw the currently collected keys.


sirlemonhead:
Thanks!

PR #568, opened 2022-02-10
Fix SDL1.2 Target Support

carnivoroussociety:
As there is still SDL 1.2 bindings currently present in the codebase this PR will enable the ability to target 1.2 again. While SDL1 has been given EoL status it is still useful for porting to older platforms such as the original Xbox or SGI systems.


sirlemonhead:
Thanks!

Issue #569, opened 2022-02-11
Ceiling scrolling reversed


Bloody-Tom:
When a ceiling texture is scrolling an it's not made "relative" in mapedit, the scrolling is reversed. In DOS ist stays in the direction its supposed to scroll. I edited E2M6 for that purpose (in mapedit, not in xmapedit (to exclude a potential problem with xmapedit))

Take  a look:
![NBlood 2022-02-11 04-52-09](https://user-images.githubusercontent.com/99460972/153535933-c2f385c9-3c3c-410e-9a5b-4e2ee2e0fd18.gif)

And for reference here the same map in Dosbox (sorry for the crappy framerate):
![blood_000](https://user-images.githubusercontent.com/99460972/153539995-eac4e88c-67ac-42de-8ce5-fc08fc6cf957.gif)

carnivoroussociety:
Hi, can you upload the modified level?

Bloody-Tom:
Here you are:
[https://www.mediafire.com/file/pm9wedjfqvnbm93/TEST.MAP/file](url)

carnivoroussociety:
Thanks, I've submitted a PR that fixes this.

sirlemonhead:
Thanks for reporting! Fixed in PR #570

PR #570, opened 2022-02-11
Blood: Fix ceiling panning angle

carnivoroussociety:
This PR fixes ceiling panning direction when not using the relative flag as seen [in this issue](https://github.com/nukeykt/NBlood/issues/569).

### DIFF
--- a/source/blood/src/sectorfx.cpp
+++ b/source/blood/src/sectorfx.cpp
@@ -298,9 +298,9 @@ void DoSectorPanning(void)
                 if (pSector->ceilingstat&64)
                     angle -= 512;
                 int xBits = (picsiz[nTile]&15)-((pSector->ceilingstat&8)!=0);
-                px += mulscale30(speed<<2, Cos(angle))>>xBits;
+                px += mulscale30(speed<<2, Cos(-angle))>>xBits;
                 int yBits = (picsiz[nTile]/16)-((pSector->ceilingstat&8)!=0);
-                py -= mulscale30(speed<<2, Sin(angle))>>yBits;
+                py -= mulscale30(speed<<2, Sin(-angle))>>yBits;
                 pSector->ceilingxpanning = px>>8;
                 pSector->ceilingypanning = py>>8;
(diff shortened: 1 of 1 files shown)


source/blood/src/sectorfx.cpp
```diff
@@ -296,11 +296,11 @@ void DoSectorPanning(void)
                 int px = (pSector->ceilingxpanning<<8)+pXSector->ceilXPanFrac;
                 int py = (pSector->ceilingypanning<<8)+pXSector->ceilYPanFrac;
                 if (pSector->ceilingstat&64)
-                    angle -= 512;
+                    angle += 512;
```
zZeck, line 299: These angles based around 2048 = 360 degrees like the [actor ones](https://wiki.eduke32.com/wiki/Ang)? This means add 90 degrees? Just curious.

carnivoroussociety: I think it is, and I believe ceiling was meant to have an inverted direction compared to floor panning but that got mixed up along the way. With relative motion the panning matches DOS, so I don't think anyone noticed this.

source/blood/src/sectorfx.cpp
```diff
@@ -298,9 +298,9 @@ void DoSectorPanning(void)
                 if (pSector->ceilingstat&64)
                     angle -= 512;
                 int xBits = (picsiz[nTile]&15)-((pSector->ceilingstat&8)!=0);
-                px += mulscale30(speed<<2, Cos(angle))>>xBits;
+                px += mulscale30(speed<<2, Cos(-angle))>>xBits;
```
zZeck, line 301: This makes more sense to me now. Is this logically the same as what you had, or was it bugged?

carnivoroussociety: Yeah, I had down wrong before. It fixed the test case in the issue but with other levels such as E2M1 the water was inverted. Turns out the calculations were fine, just ceiling inverts the angle.

PR #571, opened 2022-02-19
Blood: Clear AI Extra struct on aiInit()

carnivoroussociety:
This PR will set `aiInit()` to clear arrays `cumulDamage`, `gDudeSlope` and `gDudeExtra`. It will also restore the original vanilla behavior of leaving `pDudeExtra` uninitialized for edge case demo compatibility (spawn triggered enemies may inherit stats from previously free'd sprites).

### DIFF
--- a/source/blood/src/ai.cpp
+++ b/source/blood/src/ai.cpp
@@ -1569,6 +1569,9 @@ void aiProcessDudes(void) {
 }
 
 void aiInit(void) {
+    memset(cumulDamage, 0, sizeof(cumulDamage));
+    memset(gDudeSlope, 0, sizeof(gDudeSlope));
+    memset(gDudeExtra, 0, sizeof(gDudeExtra));
     for (int nSprite = headspritestat[kStatDude]; nSprite >= 0; nSprite = nextspritestat[nSprite]) {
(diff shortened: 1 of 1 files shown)


zZeck:
> It will also restore the original vanilla behavior of leaving `pDudeExtra` uninitialized for edge case demo compatibility (spawn triggered enemies may inherit stats from previously free'd sprites).
Maybe include a bit more in the comment about that so people know why? Could see somebody trying to fix it, or just being confused.

carnivoroussociety:
Sure, so originally before the dude extra structs were documented `aiInitSprite()` only ever reset `clock`/`teslahit` variables. I later cleared variables `sfx_priority`/`thinkTime`/`active` just for the of completion sake without considering vanilla behavior. So if there are respawning enemy such as spiders theoretically they can inherit the last active `sfx_priority`/`thinkTime`/`active` used by that sprite extra's index. This commit ensures that the original undefined behavior is kept.

As for `cumulDamage`/`gDudeSlope`/`gDudeExtra` I've cleared that based on what BloodGDX does with `aiInit()`. I encountered a demo desync because `gDudeSlope` wasn't cleared on map load, so yes this should be cleared upon level start.

Issue #572, opened 2022-02-19
Blood: Random game crash


Bloody-Tom:
In the new episode "Trauma Therapy: Episode II" in the map tt19.map looking at a specific point in a specific angle makes the game crash. It freezes, the sound keeps playing and then it crashes. No log entry is available, so I have no clue why the crash happenes. 

https://user-images.githubusercontent.com/99460972/154814067-2d257fc0-9b95-4a1e-bad5-7f5e3a81899e.mp4

The new episode can be found here:

https://www.moddb.com/mods/blood-trauma-therapy/downloads/trauma-therapy-episode-ii

carnivoroussociety:
It's a glitch with the polymer engine where it gets stuck in an infinite loop rendering room over room. I suggest using the software renderer until this is fixed.

PR #574, opened 2022-02-20
Blood: Limit ROR render to 32 times

carnivoroussociety:
This PR adds a check to prevent infinite ROR rendering as described in issue https://github.com/nukeykt/NBlood/issues/572. While does not not fix the core issue present in the polymer rendering pipeline it does stop the engine from softlocking when attempting to render infinite rooms.


sirlemonhead:
Thanks!

PR #579, opened 2022-03-02
Blood: Fix rotating platform view overflow

carnivoroussociety:
This PR fixes the view angle overflow issue when riding rotating platform sectors. For example the rotating platform on level E6M4 will cause the view to invert for a single frame due to an extreme interpolated overflow (eg: 2 -> 2023).


sirlemonhead:
Thanks!

PR #580, opened 2022-03-02
Blood: Initialize viewbob members on player spawn

carnivoroussociety:
This PR will reset view/sway members for player initialization function `playerResetPosture()`. Currently if the player is jumping/moving during a level change trigger (such as E1M3) it'll retain the view bob/sway values for the next level, causing the player's view to 'jiggle' on spawn. This PR will initialize the view bob/sway values to 0 on level change and multiplayer spawn reset.


sirlemonhead:
Cheers!

PR #582, opened 2022-03-07
Blood: Fix tesla cultist prone state

carnivoroussociety:
This PR fixes the state bug for tesla cultists. By default when a tesla cultist goes into prone state they'll fire indefinitely and not break out of it. By setting the think function reference this will force the cultist to stand back up and continue functioning.


sirlemonhead:
Thanking you!

PR #583, opened 2022-03-09
Blood: Check enemy type when adding to enemy level count

carnivoroussociety:
This PR adds additional checks to the kill manager class. When adding an enemy it will apply the same check ruleset as CountTotalKills(). This fixes a off by one miscount when blacklisted sprite types are spawned via level trigger or respawned when level is using respawning enemies setting.


sirlemonhead:
Thanks!

PR #586, opened 2022-03-11
Fix videoSetGamma() for SDL1 target

carnivoroussociety:
This PR fixes SDL 1 target for videoSetGamma(). Due to the non-bracketed if statement it would not properly set the gamma on first launch for SDL 1 builds.


sirlemonhead:
Thanks!

PR #587, opened 2022-03-11
Blood: Rename functions

carnivoroussociety:
This PR renames functions `sub_41250()`/`sub_3A158()` (names taken from Raze) and cleans up the callback for spent bullet casings. I have confirmed with demos that this PR does not change any behavior with the original game.


sirlemonhead:
Thanks!

PR #588, opened 2022-03-11
Blood: Dynamically adjust episode menu items positions

carnivoroussociety:
This PR calculates unique positions for each episode within the single player episode menu. It dynamically repositions each item so it is center weighted while the user map option is pushed down to the bottom. I've attached a gif of each possible layout from 1 episode to 6.

example


sirlemonhead:
Thanks!

PR #590, opened 2022-03-15
Blood: Fix intro playback video

carnivoroussociety:
This PR adds additional memory checks to `credLogosDos()` as well as fixing the glitched palette bug that occurs when using an OpenGL renderer (polymost/polymer). For the second intro video playback it'll display one glitched frame. This issue was introduced in commit https://github.com/nukeykt/NBlood/commit/869d85ea10f69b037b51d5aeb5f0232a73379a5a.

### DIFF
--- a/source/blood/src/credits.cpp
+++ b/source/blood/src/credits.cpp
@@ -103,7 +103,8 @@ void credLogosDos(void)
         }
     }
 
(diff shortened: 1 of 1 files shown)


sirlemonhead:
Thanks!

PR #591, opened 2022-03-16
Blood: Fix possible memory leak in credPlaySmk()

carnivoroussociety:
This PR fixes a bug introduced in commit https://github.com/nukeykt/NBlood/commit/5815e24a142d0adf92832c059889c52af5822c0a and adds a resolution check to video playback (0 pixel videos will safely exit).


sirlemonhead:
Thanks!

PR #592, opened 2022-03-16
Blood: Store net_address/net_port as CVARs

carnivoroussociety:
This PR reimplements the PR https://github.com/nukeykt/NBlood/pull/562 as CVARs. Using OSD it'll automatically save and load the last used network port and address on load.


sirlemonhead:
Thanks!

PR #593, opened 2022-03-16
Blood: Fix map scroll not using forward/backwards input

carnivoroussociety:
This fixes 2D map mode not using forwards/backwards input while in scroll mode, which matches the same behavior as 1.21 DOS.


sirlemonhead:
Ta!

PR #595, opened 2022-03-19
Blood: Add self check for burning cultist fix

carnivoroussociety:
This PR fixes a possible issue with spawned kDudeBurningCultist. As their inittype would be already set to kDudeBurningCultist, the current fix breaks as it'd constantly attempt to reset back to inittype which is kDudeBurningCultist. This resolves the issue by checking if the current sprite type is different to the inittype.


sirlemonhead:
Thanks!

Issue #610, opened 2022-07-22
[NBlood] game crashes without error message when MAP CRC does not match


Die4Ever:
initprintf("Map File does not match CRC");

I extracted E1M1.MAP, it worked fine, then I used a hex editor to change the last byte of the file to 0 and now the game crashes but there's no message like that in nblood.log

https://github.com/nukeykt/NBlood/blob/master/source/blood/src/db.cpp#L1243-L1245

Die4Ever:
so how do I see those logs?

CommonLoon102:
You can't. But if you add just one line of code then you will see it in `nblood.log`.

`initprintf()` in `baselayer.cpp`:
```
    initputs(buf);
    LOG_F(INFO, buf); // Add this line
    Xfree(buf);
```

PR #611, opened 2022-07-22
initprintf also log to nblood.log (issue #610)

Die4Ever:
I'm not sure if this would be too verbose but it could probably help people file bug reports or diagnose any issues

https://github.com/nukeykt/NBlood/issues/610

### DIFF
--- a/source/build/src/baselayer.cpp
+++ b/source/build/src/baselayer.cpp
@@ -70,7 +70,7 @@ int initprintf(const char *f, ...)
     } while ((unsigned)len >= size);
 
     initputs(buf);
-    LOG_F(INFO, buf); // also log to nblood.log
+    LOG_F(INFO, buf); // also write to the .log file
     Xfree(buf);
 
     return len;
--- a/source/build/src/baselayer.cpp
+++ b/source/build/src/baselayer.cpp
@@ -70,6 +70,7 @@ int initprintf(const char *f, ...)
     } while ((unsigned)len >= size);
 
     initputs(buf);
+    LOG_F(INFO, buf); // also log to nblood.log
     Xfree(buf);
 
     return len;


CommonLoon102:
Since this is in Build and not only for Blood, the comment is not precise because it is not necessarily just `nblood.log`. Other games log to different filenames.

Die4Ever:
> Since this is in Build and not only for Blood, the comment is not precise because it is not necessarily just `nblood.log`. Other games log to different filenames.

fixed that comment, this pull request can be squash merged to avoid the pointless extra commit

NY00123:
I've asked about this, and TerminX has confirmed my assumption. Basically, `initprintf` is supposed to be used only for messages printed to the startup window. In order to log to a file, you can use either one of the newer logging functions, or `OSD_Printf`.

In the case of NBlood's map crc check, the check itself exists in the relevant DOS version; I don't know if the use of `initprintf` was a mistake or not.

Die4Ever:
> I've asked about this, and TerminX has confirmed my assumption. Basically, `initprintf` is supposed to be used only for messages printed to the startup window. In order to log to a file, you can use either one of the newer logging functions, or `OSD_Printf`.
> 
> In the case of NBlood's map crc check, the check itself exists in the relevant DOS version; I don't know if the use of `initprintf` was a mistake or not.

Maybe you would be able to see the message if you're starting a dedicated server?

So we should probably instead fix Blood's dbLoadMap function to make it log to both? Or we should make a new function in db.cpp that logs to both, call it like mapLoadLog or something

NY00123:
I heard a bit more. Fixing the NBlood code indeed looks like the right answer.

`initprintf` calls `initputs`, which used to call `OSD_Puts` in older version of EDuke32 (and JFBuild). So in older version of NBlood, the call to `initprintf` actually did print the message to file. Maybe this was done like that so not all calls to `initprintf` (used after game startup) would have to be replaced with `OSD_Printf`.

With the transition to the newer logging functions (using loguru), this was changed, due to concerns like the ability to print things from different threads.

sirlemonhead:
Thanks!

NY00123:
The changes have been reverted for this specific PR. See the comments written in this PR earlier; The change should be in the Blood tree instead of the Build Engine.

PR #635, opened 2022-08-31
Fix header collision with LLVM 13.0.0

pkubaj:
I'm using LLVM 13.0.0 on FreeBSD 13.1-RELEASE. Error:
```
source/build/src/lz4.c:675:36: error: use of undeclared identifier 'LZ4_STREAMSIZE'
int LZ4_sizeofState(void) { return LZ4_STREAMSIZE; }
                                   ^
1 warning generated.
source/build/src/lz4.c:1426:23: error: use of undeclared identifier 'LZ4_STREAMSIZE'
    LZ4_STATIC_ASSERT(LZ4_STREAMSIZE >= sizeof(LZ4_stream_t_internal));    /* A compilation error here means LZ4_STREAMSIZE is not large enough */
                      ^
```
This is because there already is lz4.h in /usr/local/include from liblz4 package. Renaming NBlood's lz4.h to my_lz4.h and correcting include directive in lz4.c fixes the build.


sirlemonhead:
Hi pkubaj,

As this is a change to the build engine part of the project, I asked the EDuke32 developers. They stated that this doesn't seem right and 'it shouldn't be finding headers that exist in paths in the local source tree inside system directories like that'

Can you please check that you're not 'inserting /usr/local/include into his include path out of order, like after the paths specified in our makefiles' ?

PR #636, opened 2022-09-01
Blood: fix building without NOONE_EXTENSIONS

BSzili:
This PR restores the ability to build without NOONE_EXTENSIONS.


sirlemonhead:
Thank you! :)

PR #637, opened 2022-09-14
Blood: Use tick constants

tmyqlfpir:
This replaces a few magic number variables with the tick rate constants


sirlemonhead:
Thanks!

PR #638, opened 2022-09-14
Blood: Use damage constants

tmyqlfpir:
This PR renames dude info struct member at70 to curDamage, which holds the currently shifted damage values for the game session, as well as using constants to access curDamage array.


sirlemonhead:
Thanks!

PR #639, opened 2022-09-14
Blood: Tweak ROR sprite cstat behavior

tmyqlfpir:
This PR changes the behavior used for viewing player sprites across ror sectors, such as when transitioning between ror. After comparing with 1.21 on a custom level it should appear to be more accurate now.


sirlemonhead:
Thank you!

PR #640, opened 2022-09-14
Blood: Use posture constant enums

tmyqlfpir:
This PR replaces a few magic numbers with posture constraint enums.


sirlemonhead:
Thanks!

PR #641, opened 2022-09-14
Blood: Fix no monsters mode breaking on saving/loading game

tmyqlfpir:
This PR fixes the way no monsters mode is executed, and skips player sprites. Without this fix save games made without monsters will crash when attempting to load.


sirlemonhead:
Thank you again!

PR #642, opened 2022-09-14
Blood: Implemented joystick support

tmyqlfpir:
This is a very big PR, but it does work with the two controllers I've tried with, and from a friend's brief testing.

Using dynamic menu objects, the new joystick code will construct the button input pages, as well as a scaling axis menu. I used shadow warrior as my reference for implementing the joystick code, and added a feature to invert each axis. However no controller defaults are provided, as I was not too sure if this was a hold over from a wii port, or if they were broken.

Hooks have been added to game menu class, demo mode input handling and cutscenes/level end logic.

### DIFF
--- a/source/blood/src/_functio.h
+++ b/source/blood/src/_functio.h
@@ -231,12 +231,13 @@ static const char * mouseclickeddefaults[MAXMOUSEBUTTONS] =
    {
    };
 
-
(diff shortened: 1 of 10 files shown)


sirlemonhead:
Thanks! :)

Issue #612, opened 2022-07-23
[NBlood] Ambient sound discrepancy with DOS


tmyqlfpir:
For the [introduction level of this mod](https://www.moddb.com/games/blood/addons/gods-the-epic-trilogy), the ambient audio is far too loud to the point where it is clipping. This behavior is inaccurate compared to DOS, where the audio levels are within acceptable levels. Comparing with BloodGDX, it appears to match DOS volumes for both the introduction level, and E2M2's cave entrance.

I've attached two videos of NBlood (r13229) and DOS (v1.21) playing through the first level.

**Volume warning!**

https://user-images.githubusercontent.com/80724828/180590624-cf9c0adb-60fc-4c97-a9da-40e7c76c1e3b.mp4

https://user-images.githubusercontent.com/80724828/180590639-24666a81-981e-40a1-be6b-1e998dff510f.mp4

tmyqlfpir:
Fixed by [clamping channel volume for ambient audio to 127](https://github.com/tmyqlfpir/NotBlood/commit/dcfdc9ab39da839b13622f3ea390d872019cc2a5). This may have been done by the audiolib for 1.21 DOS binary.

PR #645, opened 2022-09-14
Blood: Fix incorrect ambient vol clipping channel

tmyqlfpir:
This PR fixes the issue described here https://github.com/nukeykt/NBlood/issues/612


sirlemonhead:
Thank you! :)

PR #646, opened 2022-09-14
Blood: Fade background on menu opening

tmyqlfpir:
This PR will make the background fade gradually over time on first launch, and fade to fullbright when the menu has closed.


tmyqlfpir:
Timing is constant and does not use delta timing, closing for now.

PR #650, opened 2022-09-14
Blood: Use pack constants and misc fixes

tmyqlfpir:
This PR defines a new set of kPack constants used by player inventory, as well as resolving the annoying reverb bug by reloading a save while reflective shots powerup is active. Other edits include giving crystal ball tile a constant define, and tweaking audio pitch adjustment for doppler calculation to match 1.21 behavior.


nukeykt:
Thanks!

PR #670, opened 2022-12-05
gModernMap:

NoOneBlood:
- Do not allow to repeat action for all generators if state is 0
- Allow picking random incarnation for kModernDudeCustom
- Always kill all events if object is not part of event causer trigger sequence
- Fix incorrect argument for condition
- Stop dude patrol if there is no path markers to follow (no dassert)
- Fix potentially unsafe sprite removing from dynamic sprite lists
- Rewrite initialization

### DIFF
--- a/source/blood/src/actor.cpp
+++ b/source/blood/src/actor.cpp
@@ -5849,7 +5849,7 @@ void actProcessSprites(void)
             pDb = pList->Last(); len = pList->Length();
 
             // trigger sprites from impact list
-            while (--len >= 0)
+            while (--len >= 0 && *pDb != kListEndDefault)
             {
                 pSpr = &sprite[*pDb];
                 if (!(pSpr->flags & kHitagFree))
--- a/source/blood/src/aiunicult.cpp
+++ b/source/blood/src/aiunicult.cpp
@@ -1250,14 +1250,23 @@ void killDudeLeech(spritetype* pLeech) {
 }
     
 XSPRITE* getNextIncarnation(XSPRITE* pXSprite) {
-    for (int i = bucketHead[pXSprite->txID]; i < bucketHead[pXSprite->txID + 1]; i++) {
-        if (rxBucket[i].type != 3 || rxBucket[i].index == pXSprite->reference)
-            continue;
-        
-        if (sprite[rxBucket[i].index].statnum == kStatInactive)
-            return &xsprite[sprite[rxBucket[i].index].extra];
+    
+    RXBUCKET* pRx; int s, e, t = -1;
+    getRxBucket(pXSprite->txID, &s, &e, &pRx);
+    while (s < e)
+    {
+        if (pRx->type == OBJ_SPRITE && sprite[pRx->index].statnum == kStatInactive)
+        {
+            t = pRx->index;
+            if (nnExtRandom(0, 6) == 3)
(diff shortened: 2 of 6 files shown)


CommonLoon102:
Why don't you have all these bullet points in separate commits or even PRs? That would make it much easier for other ports to integrate these changes.

NoOneBlood:
> Why don't you have all these bullet points in separate commits or even PRs? That would make it much easier for other ports to integrate these changes. Also, if there is a bug, it could be found much easier via git bisect.

Sometimes i don't see any point to commit a thing that is not yet done, and sometimes i just forget. Sorry

mjr4077au:
I have to be honest, these recent changes won't likely make it into Raze unless we get a direct contribution from you. We've made a lot of architectural changes to Blood and taking smaller commits and adapting them is fine, taking 1500 line commits is almost an impossibility.

You really need to start committing things action by action:
* Moved all conditional code to new file
* Change all the cuddled curly braces to be on a new line
* Changed this
* Changed that, etc

Additionally there's random changes to some value types or the values that need to be met for an action to occur. I get all the features are likely under development but you can't just break the API like that. These changes get lost in the noise of your huge commits, they need to be explicit commits.

There's also some changes here that frankly aren't good. Moving a lot of variables to a global state to minimise the size of function interfaces isn't great. Pass a structure around by reference or pointer if you need to but the global state can be modified anywhere by anything.

All of the `register` variable declarations are noisy as well because that's been deprecated since at least C++11. The compiler will always decide what goes into a register or what goes into the stack.

Hendricks266:
All of those are good suggestions under general circumstances.

I agree that commits in a patchset should be cleanly separated by semantic meaning. However, our team is familiar with the development cost of strict organization, especially for a volunteer like NoOne. We don't want anyone to be paralyzed by formality and structure. He has done a good job. He agreed to clearly delineate his work under the `NOONE_EXTENSIONS` preprocessor macro and made the change quickly. This level of in-source documentation gives him the freedom to iterate at his own pace with a clear test that the rest of the port is operating correctly. I have faith he will pay attention to all of our feedback regarding good practices.

I agree that implementation details can be worth improving. Globals are annoying for many reasons, that I have dealt with extensively in my employment, but in context of how prevalent they are in the entire Build ecosystem, I am not worried about them enough to obstruct development.

As for API breaks, my understanding is that these are still in development and undergoing rapid change. Advanced users of the new capabilities are in close contact with NoOne.

I would be glad to merge this as-is, though I would prefer that the `register` keyword be cleaned up first.

mjr4077au:
Thanks for replying to me. I agree with your comments as well, particularly around getting caught up in semantics that can ultimately turn contributions away, as well as the efforts he's put in. Spill's Eviction mod is second to none and its great to see Blood evolving after decades of being stuck with DOSBox.

It probably should be highlighted again that a lot of my comments come from Raze's standpoint, which of course doesn't have a lot of bearing for NBlood. We're trying our best to maintain a level of cross-port compatibility for the good of the player and now more than ever it requires smaller, digestible commits for us to manage. There's also an element of frustration as I've mentioned this to NoOne previously in passing.

Next time I'll likely voice my concerns directly rather than on another port's PR, I only did so once CommonLoon102 opened the floor for me, in a way.

PR #685, opened 2023-01-14
[NBlood] Cutscene volume fix #684 

### LINKED ISSUE #684: [NBlood] Very low cutscene volume
**Build :**
January 5 2023 - r13373

**Description:**
When playing any cutscene in the game, the cutscene audio is very low (especially when compared to Caleb's voice when starting the first level), making it hard to hear anything if your volume isn't set very high. 

**Stuff I tried out:**
I tested to see if it was an issue with my computer or my .wav files, but there doesn't seem to be anything wrong with either, since this also happens on my laptop, and the .wav files work fine outside the source port. I tried to switch out the .wav files for .ogg files, but the problem still occurs. I also tried out older versions of NBlood, thinking maybe it was a regression, but r2431 from March 17 2020 and r12727 from January 6 2022, behaved in the same way.

Here's footage, just to showcase the issue.
https://user-images.githubusercontent.com/72892666/211316233-5cdf628b-9155-4202-8d14-2a914ad15258.mp4

Oplexitie:
This PR fixes the volume in the cutscenes #684  . The _int nVolume_ in _sndStartSample_ was being multiplied by 80, while it wasn't in _sndStartWavDisk_, leading to the cutscene volume being very low.

### DIFF
--- a/source/blood/src/sound.cpp
+++ b/source/blood/src/sound.cpp
@@ -375,6 +375,8 @@ void sndStartWavDisk(const char *pzFile, int nVolume, int nChannel)
         pChannel = &Channel[nChannel];
     if (pChannel->hVoice > 0)
         sndKillSound(pChannel);
+    nVolume *= 80;
+    nVolume = clamp(nVolume, 0, 255); // clamp to range that audiolib accepts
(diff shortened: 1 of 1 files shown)


Hendricks266:
Thanks for the fix!

PR #687, opened 2023-01-16
PK3's around 45% smaller

Oplexitie:
This PR reduces the pk3 file size for nblood and dn64, with no image degradation.

nblood.pk3 (56% reduction)
178ko -> 79ko

dn64widescreen.pk3 (24% reduction)
98ko -> 74ko

Admittedly a bit pointless, but I like to do this to my images, and I wanted to share it.


Hendricks266:
Good call, merged.

PR #689, opened 2023-01-17
Blood: Pass EVENTs by const reference when possible

BSzili:
Most of the functions taking events don't poke into them, so they can be passed by const reference. This saves some work of copying them onto the stack for every call.


Hendricks266:
Nice observation, thanks!

PR #690, opened 2023-01-21
Blood: cache tilesiz values in local variables in VectorScan

BSzili:
This is a mostly cosmetic change to make the hitsprite part of VectorScan similar to the hitwall one, which caches the tilesiz values in local variables.


nukeykt:
Thanks 👍

Issue #666, opened 2022-11-28
[NBlood] Report missing Blood.rff


Phredreeke:
When NBlood is run without Blood.rff, it reports that Blood.pal is missing. While this is technically correct, it's somewhat confusing for the user as the file is normally not found outside of Blood.rff. It would IMO be helpful if the error message reported that Blood.rff was missing, since this is the file that the user is expected to see

MexMeRcUrY:
This is not limited to BLOOD.RFF but also GUI.RFF, SOUNDS.RFF

MexMeRcUrY:
creating a pull request but I'm rusty LOL
```
void Resource::Init(const char *filename)
{
    RFFHeader header;
#ifdef USE_QHEAP
    dassert(heap != NULL);
#endif

    if (filename)
    {
        handle = kopen4loadfrommod(filename, 0);
        if (handle == -1)
        {
            const char *e = "File not found ";
            char error[100];
            strcat(error, e);
            strcat(error, filename);
            ThrowError(error);
        }``
```

MexMeRcUrY:
![image](https://user-images.githubusercontent.com/379976/211143970-cfe5ac0c-936b-48f4-ae55-9369324fcab4.png)
![image](https://user-images.githubusercontent.com/379976/211143955-bda7a1ce-8672-497c-ba94-830f86f04b0a.png)
![image](https://user-images.githubusercontent.com/379976/211144017-0b64ef21-e034-4751-ba14-2cb59e0de29e.png)

PR #691, opened 2023-01-28
Resource error handing #666

MexMeRcUrY:
improve file not found error handling


CommonLoon102:
If filename can be longer than 84 chars then the program will crash or even trying to open a file with a specific filename can hack your computer.
Usually on filesystems the filenames can be about 255-260 chars.

Use a larger buffer (let it be 300 chars instead of 100).

And use `strncat` instead:
https://www.tutorialspoint.com/c_standard_library/c_function_strncat.htm

Jan200101:
Wouldn't `PATH_MAX` be good for this?

nukeykt:
Just use `ThrowError("File not found %s", filename)`

Issue #699, opened 2023-03-10
Rednukem: Keypresses are queued while awaiting verification to overwrite save file...


ExtraDrowsy:
This is annoying and happens in EDuke32 as well. When you're in the save file menu in Rednukem, if you select a save file that already exists (to overwrite) it, you're asked whether you want to overwrite and it asks you to press "y" or "n"  (`MenuType_t.Verify`). While this menu screen is displayed, it doesn't show it but your keypresses are being queued so when you finally press "y", all the keys pressed get dumped into the save file name.

Steps to reproduce:
1. Get in game in RR or DN3D.
2. Save your game (skip if you already have a save file).
3. Go and try to save again and select that file to overwrite the save.
4. When menu asks you to verify you want to overwrite by pressing "y", type a word like "foo" (or don't), then press "y".
5. When the caret appears to allow you to change the save file name, whatever you typed ("fooy" or "y") is immediately appended to the save file name.
6. Also there's another related issue where if on step 4 if you press "enter" instead of "y", it confirms, saves and takes you back to the game without having allowed you to change the save file name at all, caused by the same issue.

This is annoying because if you press "y" like the screen asks you to, your save file name gets a "y" appended to it, so if you want your save file name to remain unchanged, you then have to hit backspace to remove it, then press "enter".

I suspect most people don't run into this because they don't press "y", they just press "enter". Pressing "enter" masks this issue by saving immediately and returning you to the game without even asking you to type in a file name, this seems like a problem too, players should be given the chance to change the file name whether they press "enter" or "y". I think what's going on here is pressing "enter" instead of "y", results in "enter" being registered as soon as the caret appears, and "enter" while the caret is there would confirm your save file name and close the menu, so all of that is happening so quickly that it gives the illusion that it's not asking you to type in the save file name at all when it actually is...

Hendricks266:
I believe this is fixed by ef5f15e456e70a9a65a4623795b74a353c1157aa and 18df09f03f7bf8697f30a1db53b6c072da45a537.

PR #700, opened 2023-03-10
Rednukem: Start text input when save file overwrite verified "Y"

ExtraDrowsy:
Fix for issue #699.

This fixes the main issue and also fixes the enter key not allowing users to type a save file name on overwrite (since they have the same cause).


Hendricks266:
Thanks for the submission!

We had a separate patch merged for this upstream, which I've replicated in Rednukem as of 18df09f03f7bf8697f30a1db53b6c072da45a537. Could you confirm that your issue is resolved?

PR #712, opened 2023-05-04
Blood: Various tweaks

tmyqlfpir:
This is a collection of tweaks/fixes for Blood. They include fixing the unused implementation of the detail sprite option, attempting to load cutscene files from movie directory as per install instructions, and renaming/redefining variables. Also included is improving the shadow position calculation as it'll now test for fake floor objects.


nukeykt:
thanks

PR #722, opened 2023-06-09
Blood: Various tweaks

tmyqlfpir:
This is a collection of tweaks/fixes for Blood. They include fixing vsync so it applies during initialization, documenting CViewMap class, and fixing the selected fps setting in the menu for 144hz displays.


Hendricks266:
Thanks!

PR #729, opened 2023-07-01
Blood: Fix broken sprites in E2M4

tmyqlfpir:
This PR fixes flat enemies cstat for levels such as E2M4's enemies outside of the fireplace room.


CommonLoon102:
Might be a stupid question because I don't understand a thing, but shouldn't it be `&& !bVanilla`?

CommonLoon102:
This is a bug in the map and not in the code.
Personally, I wouldn't fix this in the code, or not like this.
Custom maps might depend on this "feature".
If we want to fix a bug in a .map file by hard-coding, shouldn't we want to hard-code the map's filename as well?

PR #735, opened 2023-08-02
Blood: Interpolate angle for 2d map view

tmyqlfpir:


Hendricks266:
Thanks! Do you have Discord? Add me, my account name is the same as here.

PR #736, opened 2023-08-02
remove [HEAD] from HUD

CommonLoon102:
Before:
image

After:
image


CommonLoon102:
Not needed since the root cause has been fixed in https://github.com/nukeykt/NBlood/commit/206fd0229015747a59d4f5c35e056d4bcaa07544.

PR #739, opened 2023-08-05
Blood: Improve prediction logic for above 60fps

tmyqlfpir:
Fixes severe unplayable judder when playing multiplayer above 60fps. This will run every game tick instead of checking when player has moved/turn.


nukeykt:
thanks

PR #741, opened 2023-08-05
Capitalized the P in "Passage", and optional DW.ini desc

BreakinBenny:
NotBlood did this already, but... consistency doesn't hurt, right? If you turn down adding a name for Death Wish, I won't complain at all...


BreakinBenny:
I expected as such.

BreakinBenny:
The latest commit in this does not give DW.INI a name, but retains the P's capitalization in "Passage".

PR #775, opened 2023-12-28
Blood: Expose quick save/load functionality as buttons

tmyqlfpir:
This PR will add two new buttons to the list of inputs, and default bind F6 and F9 to quick save and quick load, thus allowing controller players to use a function that is semi-required to enjoy Blood on difficulties above lightly broiled. It will also leave the F6 and F9 hardcoded as their respective quick save and quick load keys, so users updating NBlood won't suddenly be unable to quick save and quick load.


nukeykt:
thanks 👍

PR #777, opened 2024-01-31
Blood: Fix firing state with akimbo mode for shotgun/napalm launcher

tmyqlfpir:
This fixes an erroneous weapon state while equipped with akimbo mode. The akimbo states it fixes are:
* Firing the napalm launcher with 4-2 ammo
* Lowering QAV for shotgun and napalm launcher


nukeykt:
thanks!

Issue #799, opened 2024-04-08
Nblood weapon switch animation problem


atakanbaba3161:
while ı am switching weapons, dynamite is not showing its animation. Especially when you switch flare gun to dynamite or tesla gun to dynamite ? Latest version of nblood have this issue

PR #800, opened 2024-04-08
Blood: Reset weapon state when switching weapons

tmyqlfpir:
This PR resets the weapon state when switching weapon (this is based off of [BloodGDX's logic](https://gitlab.com/m210/BloodGDX/-/blob/942c07f31c61b6b2402ec393596ac414f683bc74/core/src/ru/m210projects/Blood/Weapon.java#L1552-1555)), and fixes issue #799


Hendricks266 (commented):
Can we use the proper macro constants instead of literals?

Also, the parentheses around the `!=` statements are superfluous.

PR #816, opened 2024-05-31
Blood: Update logic used to redraw screen border and add new -s CLI argument

### LINKED ISSUE #506: NBlood - Black Status Bar Edges
When saving or loading a save in NBlood using the full status bar (the one with the gargoyles that has stone texturing to the left and right of the respective gargoyles to fit widescreen monitors), the status bar past the gargoyles will turn black. The issue is resolved by pausing and unpausing, which will restore the texture. This has no effect on gameplay, but will occur every time you save and reload a previously made save.

### LINKED ISSUE #818: [Blood] Add -s# command line parameter
There doesn't seem to be a cmd parameter to launch Blood with selected skill level, like `-s#` in Duke/RR.

tmyqlfpir:
This PR fixes issue #506 as well as fixing the bo~~a~~rders for classic renderer while powerups and stats are being drawn.

Additionally, this PR also fulfills #818

Fixes #506
Fixes #818

### DIFF
--- a/source/blood/src/blood.cpp
+++ b/source/blood/src/blood.cpp
@@ -1229,6 +1229,7 @@ SWITCH switches[] = {
     { "c", 43, 1 },
     { "conf", 43, 1 },
     { "noconsole", 43, 0 },
+    { "s", 44, 1 },
     { NULL, 0, 0 }
 };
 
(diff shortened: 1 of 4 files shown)


Hendricks266:
"boarders" in the commit message and comments is a misspelling and should be "borders". Otherwise LGTM and I will merge when fixed.

Hendricks266:
> Done, thank you.

Thank you for the pull request and quick update.

PR #819, opened 2024-06-09
Blood: Dynamically scale sky parallax drawing for FOV

### LINKED ISSUE #383: Suggestion: Mirror skyboxes when using display options that allow you to see more than intended
Not a bug or anything. Just a quality of life fix for skyboxes looking odd when ticking on "Centre Horizon Line", and changing the FOV. Example below, you can see a rather large seam where the image has repeated:

example

Thank you for the port and have a nice day.

tmyqlfpir:
This PR fixes issue #383 by dynamically scaling the FOV to ensure that the skybox drawing does not overflow. 

An example with an FOV of 120 and set to center horizon line.

Before:
blud0001
After:
blud0000

Fixes #383

### DIFF
--- a/source/blood/src/aibat.cpp
+++ b/source/blood/src/aibat.cpp
@@ -74,6 +74,19 @@ static void BiteSeqCallback(int, int nXSprite)
 {
     XSPRITE *pXSprite = &xsprite[nXSprite];
     spritetype *pSprite = &sprite[pXSprite->reference];
+    /*
+     * workaround for
+     * pXSprite->target >= 0 && pXSprite->target < kMaxSprites in file NBlood/source/blood/src/aibat.cpp at line 85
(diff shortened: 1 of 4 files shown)


Hendricks266:
I like this idea. I need to mentally run "red team" on it before merging.

Could you post the derivations for the constants you used?

Hendricks266:
Should the calculation account for the y-height of the sky tile? What if the user loads replacement sky tiles that are taller, kind of like the skybox version of widescreen weapons?

Hendricks266:
What if you did change the tile, such as to extend it upwards? What about mods, do any have new art with a different height?

Of course cubemap skyboxes are unaffected.

Hendricks266:
The calculation needs to account for tilesiz[].y of the paper sky in question.

PR #823, opened 2024-06-20
Fix rumble detection

### LINKED ISSUE #750: Can't enable controller rumble on an Xbox controller.
Looking through code i saw mentions about rumble/force feedback but i couldn't find any way to enable it in Duke or Blood. Otherwise controller support is great.

tmyqlfpir:
This PR fixes rumble detection for controllers with EDuke32.

Fixes #750

### DIFF
--- a/source/build/src/sdlayer.cpp
+++ b/source/build/src/sdlayer.cpp
@@ -974,10 +974,10 @@ void joyScanDevices()
 
                 inputdevices |= DEV_JOYSTICK;
 
-#if SDL_VERSION_ATLEAST(2, 0, 9)
-                if (EDUKE32_SDL_LINKED_PREREQ(linked, 2, 0, 9))
+#if SDL_VERSION_ATLEAST(2, 0, 18)
+                if (EDUKE32_SDL_LINKED_PREREQ(linked, 2, 0, 18))
                 {
-                    if (!SDL_GameControllerRumble(controller, 0xc000, 0xc000, 10))
+                    if (SDL_GameControllerHasRumble(controller))
                         joystick.hasRumble = 1;
                     else DVLOG_F(LOG_INPUT, "Couldn't init controller rumble: %s.", SDL_GetError());
                 }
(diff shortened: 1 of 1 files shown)


Hendricks266:
This change looks fine in itself. I would prefer if it maintained an alternate code path for the older versions of SDL2 before the rumble behavior was changed.

source/build/src/sdlayer.cpp
```diff
@@ -974,10 +974,17 @@ void joyScanDevices()
...
 
-#if SDL_VERSION_ATLEAST(2, 0, 9)
+#if SDL_VERSION_ATLEAST(2, 0, 18)
+                if (EDUKE32_SDL_LINKED_PREREQ(linked, 2, 0, 18))
+                {
+                    if (SDL_GameControllerHasRumble(controller))
+                        joystick.hasRumble = 1;
+                    else DVLOG_F(LOG_INPUT, "Couldn't init controller rumble: %s.", SDL_GetError());
+                }
+#elif SDL_VERSION_ATLEAST(2, 0, 9)
```
Hendricks266, line 984:
```suggestion
                else
#endif
#if SDL_VERSION_ATLEAST(2, 0, 9)
```

source/build/src/sdlayer.cpp
```diff
@@ -1020,10 +1027,17 @@ void joyScanDevices()
...
 
-#if SDL_VERSION_ATLEAST(2, 0, 9)
+#if SDL_VERSION_ATLEAST(2, 0, 18)
+                if (EDUKE32_SDL_LINKED_PREREQ(linked, 2, 0, 18))
+                {
+                    if (SDL_JoystickHasRumble(joydev))
+                        joystick.hasRumble = 1;
+                    else DVLOG_F(LOG_INPUT, "Couldn't init joystick rumble: %s.", SDL_GetError());
+                }
+#elif SDL_VERSION_ATLEAST(2, 0, 9)
```
Hendricks266, line 1037:
```suggestion
                else
#endif
#if SDL_VERSION_ATLEAST(2, 0, 9)
```

PR #832, opened 2024-08-02
Blood: Remove unused fps offset code

tmyqlfpir:
This PR removes some legacy code relating to fps offset, which has since been removed.

### DIFF
--- a/source/blood/src/menu.cpp
+++ b/source/blood/src/menu.cpp
@@ -80,7 +80,6 @@ void SetVideoMode(CGameMenuItemChain *);
 void SetWidescreen(CGameMenuItemZBool *);
 void SetFOV(CGameMenuItemSlider *);
 void UpdateVideoModeMenuFrameLimit(CGameMenuItemZCycle *pItem);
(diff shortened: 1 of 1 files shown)


Hendricks266:
> Blood: Add option to disable palette emulation for polymost

Our philosophy is that breaking the game's aesthetic _should_ be hidden behind cvars, and the only reason it wasn't in the past is that our old implementation was performance-intensive, before we had a shader for it.

The only exception is if a user wants texture filtering, but any such option should do the right things behind the scenes, rather than confuse them with a palette emulation toggle.

> Blood: Remove outdated fps offset code

Looks fine. This code is probably broken anyway. This functionality would be better served by a submenu that the engine provides to all games, which my intent to implement.

> Blood: Reject hardcoded hotkeys keys when binding new key in settings 

I don't agree with a key banlist. In VoidSW, I recently implemented rebindable gamefuncs for Quick Save and Quick Load. It is possible to, for example, rebind Quick Save to F5, which will replace the shortcut to the Sound Options menu.

PR #834, opened 2024-08-12
Added 2 command line options to force a version of the game

wcapes:
Added 2 command line options to force a version of the game

-ps to force using powerslave.dat
-exhumed to force using exhumed.dat

Reason: Allow rename of STUFF.DAT to either .DAT to allow both version of the games in the same folder.   Allows external launches/batch files and also no need to use the UI to select either one


sirlemonhead:
Thanks for the PR!

I think if we want to do something like this, where we can let the user choose which .dat/,grp to use for the base data, we'd want something similar to what Duke3D does with the '-gamegrp' parameter and have the utility/launcher specify the file name. We'd need to make a small change similar to your PR for that I think. And I think we'd need to check if -nosetup is available too, to skip showing the EDuke32 setup screen.

Of course, the problem then is at a .dat level, there's not a big difference between Powerslave and Exhumed's dat files from what I recall. Both have each others Publisher logo and game logo graphics. 

The two files are slightly different, however, so we do use the EDuke32 game detection to, at runtime, determine which sets of logos we display and whether the game's menu shows it as Powerslave or Exhumed.

Is there a benefit to the user being able to select between the two? From my (hazy) memory of things, I'd prefer Exhumed as it was the more feature complete of the two PC versions.

I've not looked at the code in a while, to be honest, so let me know if I'm on the wrong track here with this!

Hendricks266:
Yeah, implementing `-gamegrp` is the correct approach here, which should be straightforward to do by following what the other games with that option do.

I see value in supporting the Powerslave GRP because it lets players use their own copy of the game on CD.

@wcapes @warrencapes In the future, please follow these guidelines for your commit message format: https://cbea.ms/git-commit/

wcapes:
Good day.

Thanks for the feedback

I just did it for my personal little project I started, and thought I'd share, however I do understand everyone's opinion and approach, so happy if you want to abandon this.

sirlemonhead:
I appreciate anyone's willingness to help out :) I didn't realise the gamegrp argument was even missing so I've added it now.

So you can do stuff like:

-gamegrp exhumed.DAT -nosetup
or
-gamegrp powerslave.DAT -nosetup

on the command line to tell the game to use either of the version DAT file and go straight into the game menus.

Hendricks266:
Thanks for the suggestion, and I hope what we've come up with works for you. Let us know if we can help with anything else or review more changes.

wcapes:
I know this is closed, just want to thank you guys for jumping on this so fast - will all the engines out there I still use yours for a run of a few levels here and there :)

PR #842, opened 2024-09-05
Blood: Add option to select MIDI MME device

### LINKED ISSUE #831: NBlood no native MIDI
I'm using NBlood on Windows 11. When I set the music to native MIDI, it doesn't play.
I use VST MIDI Driver + Yamaha S-YXG50 VSTi for MIDI music. It works fine with other games which uses MIDI for soundtrack.

Sorry for my bad English grammer. it's not my native language

tmyqlfpir:
This PR allows users to change the MME device index for system MIDI playback (only for Windows based systems).

Fixes #831

### DIFF
--- a/source/blood/src/menu.cpp
+++ b/source/blood/src/menu.cpp
@@ -39,6 +39,10 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #include "screen.h"
 #include "sound.h"
 #include "view.h"
+#ifdef _WIN32 // required for MME midi device selection
+#include "driver_winmm.h"
+#include "midi.h"
+#endif
 
 void SaveGame(CGameMenuItemZEditBitmap *, CGameMenuEvent *);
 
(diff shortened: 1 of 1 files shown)


source/blood/src/menu.cpp
```diff
@@ -39,6 +39,10 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #include "screen.h"
 #include "sound.h"
 #include "view.h"
+#ifdef _WIN32 // required for MME midi device selection
+#include "../src/driver_winmm.h"
+#include "../src/midi.h"
+#endif
```
Hendricks266, line 45: I'm not a fan of this kind of relative path construction, it violates encapsulation. Instead, APIs should be added or moved to public headers.

Hendricks266: If you make the change to the headers as its own commit, I will cherry-pick it upstream.

CommonLoon102:
I see unrelated commits in this PR. Normally a new feature should be in a feature branch inside a fork, but I see here master.

Hendricks266:
> I see unrelated commits in this PR. Normally a new feature should be in a feature branch inside a fork, but I see here master.

I'm okay with slipstreaming small fixes into feature branches as long as they're cleanly separated into their own commits, which they are. I do agree that it might be better not to use the `master` branch for pull requests.

PR #843, opened 2024-09-05
Blood: Add option to show stats on automap only

tmyqlfpir:
This PR adds the option to only allow stats to appear on the automap.

example


CommonLoon102:
If I can have a suggestion: probably a 3 mode switch would be better here. On/Off/Automap Only

Hendricks266 (approved):
Looks good. I'll merge once the current build job finishes.

PR #844, opened 2024-09-06
Blood: Fix tiles broken from recompression attempts

tmyqlfpir:
This PR fixes errors for tiles 9218 and 9219 introduced in #687 and #779.

Comparing to the original tiles before these two PRs, I compared the files by using a difference filter in photoshop and increased the brightness to exaggerate the pixels that differ.

Original vs after PR #779
originalcompressed

Difference
current

Original vs after PR #779
original
compressed

Difference
difference

I went through and recompressed all tiles by using the original images before PR #687. [PNGOUT](http://advsys.net/ken/utils.htm#pngout) and [ECT](https://github.com/fhanau/Efficient-Compression-Tool) were both used.
The commands to compressed the images are as follows
```
pngout /f0 /b128 /kt /k0 /v /y
ect -9 -strip -keep --pal_sort=120 --allfilters
pngout /f0 /b128 /kt /k0 /v /y
```


Oplexitie:
Went back and checked the assets of the other pk3's just to be sure this didn't affect them.
Thankfully, this only affected the Blood assets that the op mentioned.

Spent a few hours figuring out what happened. The affected assets used [PNGQUANT](https://pngquant.org/) (a lossy image compressor), and I included them by accident. Apologies for my blunder.

PR #846, opened 2024-09-09
Blood: Add turn speed slider for keyboard input

tmyqlfpir:
This PR adds the ability to change turn speed and turn off turn speed acceleration for Blood much like BloodGDX.

turnspeed


Hendricks266:
Looking for feedback on the name of the menu option and cvar

PR #848, opened 2024-09-12
Mact: Add single axis deadzone flag for analog stick distance calculation

tmyqlfpir:
This PR adds a new flag to joysticks deadzone calculation, and helps with general analog stick movement such as the ability to walk perfectly straight when tilted fully forward on the left stick.

Originally the mact library would calculate deadzone based on the total axes input for a single analog stick instead of treating this as two independent axis for the deadzones. This is usually fine for looking left/right and up/down, however when it comes to strafing this can make it can become difficult to walk in a perfect line while also fully tilting the analog stick forward.

Here is a comparison with the setting off and on, and attempting to walk forward.

https://github.com/user-attachments/assets/60fd14b5-da82-4fb2-b20e-46d8881eb7b3

https://github.com/user-attachments/assets/219d9cd6-c68d-4c93-a296-9c0ff221fc96

New settings for axis option menu.

Untitled

### DIFF
--- a/source/blood/src/_functio.h
+++ b/source/blood/src/_functio.h
@@ -356,7 +356,7 @@ static const int32_t joystickanalogdeaddefaults[MAXJOYAXES] =
    };
 
 
-static const int32_t joystickanalogaxissolodeadzone[MAXJOYAXES] =
+static const int32_t joystickanalogdecoupledaxes[MAXJOYAXES] =
    {
    1,
    0,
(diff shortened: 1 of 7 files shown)


CommonLoon102:
off.mp4: Caleb drank too much wine

Hendricks266:
I like "Decoupled Axes" as the name for this.

PR #854, opened 2024-10-14
Blood: Fix enemy death sfx getting cut and fix cultist sfx not getting cut on death

### LINKED ISSUE #804: [NBlood] Enemy death sounds randomly cut out
Demo (watch with sound on):

[demo.webm](https://github.com/nukeykt/NBlood/assets/128177553/91059ad1-af2d-452f-8ba4-e45ef462c679)

While full audio is:

https://github.com/nukeykt/NBlood/assets/128177553/158b7b1b-d7be-49bb-84d7-b67540dd4cd6

I'm not the first one to encounter this bug: https://old.reddit.com/r/bloodfps/comments/13evddl/nblood_sound_bug/

tmyqlfpir:
This PR fixes a common issue of channels being cut out upon an enemy's death, and resolves an original game bug of cultist alerts continuing to play on death.

When an enemy is killed by a hitscan weapon such as the shotgun or tommy gun, it'll spawn a spurt of blood that has a random chance of triggering a splattering sfx upon hitting the ground - and cutting off the still ongoing enemy death sfx. This PR will set the sound effect trigger to only play if the channel is available, instead of cutting off the voice.

For cultists, when they are killed at point blank range with the shotgun, it'll trigger the activate callout sfx, then the death sfx in the same tick. This PR resolves this by setting all cultists to use their own unique channel slot for callouts, so they are cut off upon dying or getting hurt.

Fixes #804

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -811,9 +811,9 @@ blood_game_objs := \
 
 ifeq ($(NOONE_EXTENSIONS),1)
     blood_game_objs += nnextsif.cpp
-	blood_game_objs += nnexts.cpp
-	blood_game_objs += nnextstr.cpp
-	blood_game_objs += nnextcdud.cpp
+    blood_game_objs += nnexts.cpp
+    blood_game_objs += nnextstr.cpp
+    blood_game_objs += nnextcdud.cpp
     blood_game_objs += aicdud.cpp
 endif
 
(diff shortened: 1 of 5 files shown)


Hendricks266:
Looks great. Only comments: I would prefer if `Blood: Tweak aiActivateDude() nChannel variable` were squashed into `Blood: Mute cultist alert calls on death`, rather than implementing something one way and changing it later in the same patchset. Also, should the literal 16384 used for the definition of `nChannel` be some macro constant? I'm guessing MAXSPRITES.

PR #869, opened 2024-12-08
Blood: Precache baby spiders for mother spider

tmyqlfpir:
This PR adds tile caching for baby spider types when a mother spider is present in the level. This should fix invalid sprites when these enemies are ignited on fire such as on E4M2.


CommonLoon102:
Oh, then probably it won't crash anymore in co-op with respawning monsters at the second boss.

PR #884, opened 2024-12-29
Blood: Add 'pitchfork only' autoaim option

tmyqlfpir:
Smaller enemies such as the hands, rats and fish appear to be designed in mind for autoaim as it's incredibly difficult to hit them precisely when autoaim is turned off.

This PR adds a new autoaim setting to allow players to always have autoaim set for the pitchfork weapon, so players who prefer not to use autoaim can have it only applied for the pitchfork.

test

### DIFF
--- a/source/blood/src/menu.cpp
+++ b/source/blood/src/menu.cpp
@@ -431,7 +431,8 @@ CGameMenuItemChain itemOptionsChainOld("OLD MENU", 1, 0, 170, 320, 1, &menuOptio
 const char *pzAutoAimStrings[] = {
     "NEVER",
     "ALWAYS",
-    "HITSCAN ONLY"
+    "HITSCAN ONLY",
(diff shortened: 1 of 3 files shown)


Hendricks266:
My intuition would be, if I had autoaim disabled and I clearly missed but it hit anyway, that would not feel right. That said, I see that Blood enables autoaim unconditionally for the Voodoo Doll and the Life Leech.

In EDuke32, we make it so that if the crosshair is actually over the target when hitscan weapons are fired, they always hit the target regardless of random weapon spread, even if autoaim is disabled. This is because vanilla Duke 3D unconditionally uses autoaim, and the game was balanced in that way. In contrast, vanilla Blood has a setting to disable autoaim. This may be the true solution to what this PR intends to fix, and possibly even the forced autoaim in the original game.

ElJulo25:
I prefer no auto aim imho

PR #897, opened 2025-02-15
Blood: Reset palette colors on cutscene trigger

tmyqlfpir:
This addresses a bug if the user has set palette emulation to false for polymost, and tries to render FMV.

bug


Hendricks266:
I'll accept this as a workaround though ideally we wouldn't need to call `restartvid` around cinematics.

PR #898, opened 2025-03-13
rednukem: always disable unsupported polymer renderer

tmyqlfpir:
This PR forcefully disables the unsupported polymer rendering mode for Duke Nukem 64.

Fixes #740

PR #900, opened 2025-03-14
rednukem: Only disable polymer renderer for duke 64

tmyqlfpir:
Addresses overzealous correction made by PR #898


Hendricks266:
Thanks for the quick fix.

PR #915, opened 2025-05-13
Blood: Use accurate calculation for enemy/player created explosions

tmyqlfpir:
This PR is based off various improvements made to NotBlood, DOSBlood and Raze for their explosion radius calculation.

https://github.com/clipmove/DOSBlood/commit/b2835f675dc969573dcbfa41b3f71aee0414f756

By improving the scanning logic and using `getwalldist()`, it'll check midpoint of walls instead of only the start/end point of walls at the expense of additional calculations which is harmless with 2025+ era hardware.


Hendricks266:
Is there a strong reason for why `nDist16` is a variable, calculated in advance? My best guess would be some kind of pipelining trick important for DOSBlood.

Hendricks266:
Adding an extra branch is actually worse for performance. I don't see why `nDist<<4` can't just be on the line at the bottom where it is used.

PR #941, opened 2025-09-22
Blood: Move input polling at start of game loop and use delta scaling for joystick input

tmyqlfpir:
This PR addresses an issue related with #940

Joystick now uses the existing function `scaleAdjustmentToInterval()`, which is used to compensate for a dynamic tickrate.

### DIFF
--- a/source/blood/src/controls.cpp
+++ b/source/blood/src/controls.cpp
@@ -505,8 +505,8 @@ void ctrlGetInput(void)
 
     if (CONTROL_JoystickEnabled) // controller input
     {
-        input.strafe -= info.dx>>1;
-        input.forward -= info.dz>>1;
+        input.strafe -= int(scaleAdjustmentToInterval(info.dx)/2.f);
+        input.forward -= int(scaleAdjustmentToInterval(info.dz)/2.f);
         if (!run) // when autorun is off/run is not held, reduce overall speed for controller
         {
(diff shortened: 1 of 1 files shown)


source/blood/src/controls.cpp
```diff
@@ -515,12 +515,11 @@ void ctrlGetInput(void)
...
             if (gMouseAim)
-                input.q16mlook = fix16_sadd(input.q16mlook, fix16_sdiv(fix16_from_int(info.dpitch>>4), F16(128)));
+                input.q16mlook = fix16_sadd(input.q16mlook, fix16_sdiv(fix16_from_float(scaleAdjustmentToInterval(info.dpitch)/16.f), F16(128)));
             else
-                input.forward -= info.dpitch>>1;
+                input.forward -= int(scaleAdjustmentToInterval(info.dpitch)/2.f);
         }
-        if (input.q16turn == 0)
-            input.q16turn = fix16_sadd(input.q16mlook, fix16_sdiv(fix16_from_int(info.dyaw>>4), F16(32)));
+        input.q16turn = fix16_sadd(input.q16turn, fix16_sdiv(fix16_from_float(scaleAdjustmentToInterval(info.dyaw)/16.f), F16(32)));
```
Hendricks266, line 522: I want to double check that the change on this line from `q16mlook` to `q16turn` is intentional and correct. It certainly looks better this way.

PR #946, opened 2025-10-01
Blood: Cap map zoom speed to game tickrate

tmyqlfpir:
This PR limits the map zoom speed in 2D map mode for NBlood, so it zooms at same speed regardless of FPS.


Hendricks266:
I will test this and merge if it seems good.

Hendricks266:
Much better.

PR #947, opened 2025-10-03
Blood: Add interpolated sector textures

tmyqlfpir:
This PR adds interpolation to wall/ceiling/floor texture scrolling.

It also exposes a new ROR-friendly `gotsector` variable called `gotsectorROR`.


nukeykt:
can you check what happen if you cast them to uint32_t pointer?
```
uint32_t* p1 = (uint32_t*)gotsector;
uint32_t* p2 = (uint32_t*)gotsectorROR;
for (uint32_t i = 0; i < ARRAY_SIZE(gotsector) / 4; i++)
{
    *p2++ |= *p1++;
}
```

PR #948, opened 2025-10-04
Blood: Scale 2D map view to output resolution

tmyqlfpir:
This PR fixes a bug where the 2D map view does not scale to the output window size.

Below is an example of the issue in action. This is set to maximum map zoom.

### Before

before

### After

after

### DOSBOX reference (640x400)

dosbox

### DIFF
--- a/source/blood/src/map2d.cpp
+++ b/source/blood/src/map2d.cpp
@@ -216,7 +216,7 @@ void CViewMap::Draw(void)
 
 void CViewMap::Process(int nX, int nY, short nAng)
 {
-    nZoom = gZoom*((ydim/200)>>1); // scale to display resolution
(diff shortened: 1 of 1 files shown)


source/blood/src/map2d.cpp
```diff
@@ -216,7 +216,7 @@ void CViewMap::Draw(void)
 
 void CViewMap::Process(int nX, int nY, short nAng)
 {
-    nZoom = gZoom;
+    nZoom = gZoom*((ydim/200)>>1); // scale to display resolution
```
Hendricks266, line 219:
```suggestion
    nZoom = scale(gZoom, ydim, 200) >> 1; // scale to display resolution
```
I believe this gives the intended result while preserving precision and avoiding overflow.

